PR #2, opened 2015-05-20
add test for messages with static arrays

dirk-thomas:
Connects to ros2/ros2#42

Depends on ros2/rmw_connext#36.

@esteve @tfoote @wjwwood Please review.

### DIFF
--- a/test_communication/test/message_fixtures.hpp
+++ b/test_communication/test/message_fixtures.hpp
@@ -19,6 +19,7 @@
 
 #include <test_communication/Empty.h>
 #include <test_communication/Primitives.h>
+#include <test_communication/StaticArrayPrimitives.h>
 
 
 std::vector<test_communication::Empty::Ptr>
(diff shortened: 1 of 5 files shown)


tfoote:
+1

test_communication/test/message_fixtures.hpp
```diff
@@ -91,4 +92,31 @@ get_messages_primitives()
...
+    msg->uint8_values = {0, 255, 0};
+    msg->int16_values = {0, 32767, -32768};
+    msg->uint16_values = {0, 65535, 0};
+    msg->int32_values = {0, 2147483647, -2147483648};
+    msg->uint32_values = {0, 4294967295, 0};
+    msg->int64_values[0] = 0;
+    msg->int64_values[1] = 9223372036854775807;
+    msg->int64_values[2] = -9223372036854775808UL;
+    msg->uint64_values = {0, 18446744073709551615UL, 0};
+    msg->string_values = {"", "max value", "min value"};
```
esteve: Could you add a test for arrays with only one value? IIRC, GCC and Visual Studio break when parsing them unless you add double braces `{{ 42 }}`

esteve:
+1

test_communication/test/message_fixtures.hpp
dirk-thomas: Sure, the upcoming dynamic arrays will have fixtures with a single value. (But with gcc they seem to compile fine for me until now.)

esteve: That's right, it's dynamic arrays the ones that have problems with single-valued aggregate initializations.

PR #13, opened 2015-06-18
add tests for publish/subscribe in a single process

dirk-thomas:
This test currently fails with the Connext rwm implementations.


dirk-thomas:
Please review this as well as the referenced PR on `rmw_connext`.

wjwwood:
lgtm

tfoote:
+1

dirk-thomas:
This was auto-closed when the referenced PR was merged.

esteve:
+1

dirk-thomas:
Since the test binary only uses the return code to signal success / failure it needs to explicitly generate a XUnit result file: 3f0ab3f6f1389a473841bfbaa6d214d199a533e5

Issue #10, opened 2015-06-11
Add tests for Parameters


wjwwood:
Similar to the tests for pub/sub and services.

tfoote:
ongoing in this branch: https://github.com/ros2/system_tests/tree/test_parameters

tfoote:
Tests are running on opensplice and connext connext_dynamic is hanging.  I need to debug more to find out what's going wrong. It's only basic coverage at the moment.

gerkey:
From today's meeting: there is support in ament for: gtest, gmock, and nose.  @dirk-thomas will add a pointer to the relevant documentation.

gerkey:
Documentation is in the CMake code: https://github.com/ament/ament_cmake/blob/master/ament_cmake_gtest/cmake/ament_add_gtest.cmake

tfoote:
http://ci.ros2.org/job/ros2_batch_ci_linux/261/ pass
http://ci.ros2.org/job/ros2_batch_ci_osx/186/ pass
http://ci.ros2.org/job/ros2_batch_ci_windows/265/ (failing to find gtest. has anyone seen this error before?)

dirk-thomas:
If you look at any other Windows job you will find that GTest is always missing. Simply because GTest is not installed on the Windows slave. This will be resolved once we merge ament/ament_cmake#22

PR #14, opened 2015-07-07
Add tests for parameters

tfoote:
This has some basic tests. However I need some help on the RTI Connext Dynamic implmeentation tests which hang. 

How much details do we want to implement before we add support for a testing framework like gtest. Writing detailed unit tests with return values only is not very productive. 

This is currently running the tests and checking their values only. 

Connects to #10

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -135,6 +135,46 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    if(TARGET local_parameters_cpp__${middleware_impl})
+      ament_add_gtest(local_parameters_cpp__${middleware_impl} SKIP_LINKING_MAIN_LIBRARIES "test/local_parameters.cpp")
+      add_dependencies(local_parameters_cpp__${middleware_impl} ${PROJECT_NAME})
+      rosidl_target_interfaces(local_parameters_cpp__${middleware_impl}
+        ${PROJECT_NAME} ${typesupport_impl})
+      ament_target_dependencies(local_parameters_cpp__${middleware_impl}
+        "${middleware_impl}"
+        "rclcpp")
+    endif()
+
+    if(TARGET test_remote_parameters_cpp__${middleware_impl})
+      # Parameter tests single implementation
+      add_executable(test_remote_parameters_cpp__${middleware_impl} "test/test_remote_parameters.cpp")
+      target_link_libraries(test_remote_parameters_cpp__${middleware_impl} ${GTEST_LIBRARIES})
+      # the above could be replaced by a helper which won't register the gtest for running
+      # ament_add_gtest_subtest(test_remote_parameters_cpp__${middleware_impl} "test/test_remote_parameters.cpp")
+      add_dependencies(test_remote_parameters_cpp__${middleware_impl} ${PROJECT_NAME})
+      rosidl_target_interfaces(test_remote_parameters_cpp__${middleware_impl}
+        ${PROJECT_NAME} ${typesupport_impl})
+      ament_target_dependencies(test_remote_parameters_cpp__${middleware_impl}
+        "${middleware_impl}"
+        "rclcpp")
+      add_executable(test_parameters_server_cpp__${middleware_impl} "test/test_parameters_server.cpp")
+      add_dependencies(test_parameters_server_cpp__${middleware_impl} ${PROJECT_NAME})
+      rosidl_target_interfaces(test_parameters_server_cpp__${middleware_impl}
+        ${PROJECT_NAME} ${typesupport_impl})
+      ament_target_dependencies(test_parameters_server_cpp__${middleware_impl}
+        "${middleware_impl}"
+        "rclcpp")
+      set(TEST_PARAMETERS_SERVER_EXECUTABLE "${CMAKE_CURRENT_BINARY_DIR}${test_executable_subfolder}/test_parameters_server_cpp__${middleware_impl}${test_executable_extension}")
+      set(TEST_REMOTE_PARAMETERS_EXECUTABLE "${CMAKE_CURRENT_BINARY_DIR}${test_executable_subfolder}/test_remote_parameters_cpp__${middleware_impl}${test_executable_extension}")
+      configure_file(
+        test/test_parameters_server.py.in
+        test_parameters_server_cpp__${middleware_impl}.py
+        @ONLY
+      )
+      ament_add_nose_test(parameter_server_cpp__${middleware_impl} "${CMAKE_CURRENT_BINARY_DIR}/test_parameters_server_cpp__${middleware_impl}.py" TIMEOUT 15)
+        set_tests_properties(parameter_server_cpp__${middleware_impl} PROPERTIES DEPENDS "test_parameters_server_cpp__${middleware_impl} test_remote_parameters_cpp__${middleware_impl}")
+    endif()
+
   endforeach()
 endif()  # AMENT_ENABLE_TESTING
 
--- a/test_rclcpp/test/local_parameters.cpp
+++ b/test_rclcpp/test/local_parameters.cpp
@@ -1,0 +1,48 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <iostream>
+#include <stdexcept>
+#include <string>
+#include "gtest/gtest.h"
+
+#include <rclcpp/rclcpp.hpp>
+
+#include "parameter_fixtures.hpp"
+
+
+TEST(parameters, local_synchronous) {
+  auto node = rclcpp::Node::make_shared(std::string("test_parameters_"));
+  // TODO(esteve): Make the parameter service automatically start with the node.
+  auto parameter_service = std::make_shared<rclcpp::parameter_service::ParameterService>(node);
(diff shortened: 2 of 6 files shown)


dirk-thomas:
All test launch files fail with:

    NameError: name 'test_parameters' is not defined

dirk-thomas:
The Python launch file does not follow the style conventions.

test_communication/CMakeLists.txt
```diff
@@ -174,6 +182,21 @@ if(AMENT_ENABLE_TESTING)
...
+      test_parameters_cpp__${middleware_impl1}.py
+      @ONLY
+    )
+    ament_add_nose_test(parameters_cpp__${middleware_impl1} "${CMAKE_CURRENT_BINARY_DIR}/test_parameters_cpp__${middleware_impl1}.py" TIMEOUT 15)
+    set_tests_properties(
+      parameters_cpp__${middleware_impl1}
+      PROPERTIES DEPENDS "test_parameters_cpp__${middleware_impl1}"
+    )
+
+
```
dirk-thomas: Please remove the two empty lines.

test_communication/test/test_parameters.cpp
```diff
@@ -0,0 +1,87 @@
...
+#include <rclcpp/rclcpp.hpp>
+
+
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+
+  auto start = std::chrono::steady_clock::now();
+
+  auto node = rclcpp::Node::make_shared(std::string("test_parameters_"));
```
dirk-thomas: `std::string` is not necessary here.

test_communication/test/test_parameters.cpp
```diff
@@ -0,0 +1,87 @@
...
+    if (!result.successful) {
+      std::cerr << "Failed to set parameter: " << result.reason << std::endl;
+      return 1;
+    }
+  }
+
+  // List the details of a few parameters up to a namespace depth of 10.
+  auto parameters_and_prefixes = parameters_client->list_parameters({"foo", "bar"}, 10);
+  for (auto & name : parameters_and_prefixes.names) {
+    std::cout << "Parameter name: " << name << std::endl;
```
dirk-thomas: The test should actually test the received values and assert that they are what is expected. Same below.

dirk-thomas:
Instead of streaming to cout / cerr incrementally this should use printf / fprintf to avoid interleaving.

dirk-thomas:
Connext Dynamic fails with the following for me which indicate some type mismatch (may be related to upper bounds?):

    DDS_DynamicData_set_string:type mismatch for field <no name> (id=1)

dirk-thomas:
The Connext dynamic test currently hangs forever since the sync API has no timeouts. May be it would be better to cover the async API in a first test which has timeouts instead.

esteve:
@dirk-thomas I pushed a test that exposes the issue with subscriptions not being fired. The test will hang at the end because we `rclcpp::spin_until_future_complete` does not support timeouts yet.

dirk-thomas:
The current behavior is that `rmw_wait()` returns when one or multiple samples are ready to be taken. But the executor only takes one of them and then calls `rmw_wait()` again. This call won't return until *new* samples arrive.

So either rmw has to always return as long as not yet taken samples are available or the executor has to handle available sample after wait returns. Since the `SyncParametersClient` currently embeds its own executor it is not feasible to make the executor stateful at the moment.

esteve:
@dirk-thomas I have a local branch that modifies the executors so that there's a global default one that can be reused instead of instantiating a new one.

dirk-thomas:
ros2/system_tests#20 contains a test for the spinning problem.

ros2/rmw_connext#62 fixes the behavior of `rmw_connext_cpp`.

dirk-thomas:
@esteve Can you please implement a similar fix as ros2/rmw_connext#62 (and for opensplice) to change the conditions for services too? I would expect that to fix this broken test.

esteve:
@dirk-thomas I won't be able to do it today or tomorrow, so if you have the fix fresh in your mind, go for it. Thanks!

dirk-thomas:
The patch will look exactly the same way as for the subscription. Nothing new or different, just effort to put it together.

tfoote:
I've updated it to use gtest and actually test the results. It's revealed some underlying errors. The returned parameters have different values than the set parameters. 

I will next look at the implementation to try to get the unit tests passing. I triggered a CI job to verify the same behavior on the farm: http://ci.ros2.org/job/ros2_batch_ci_linux/122/

```
$ ./test_parameters_cpp__rmw_connext_cpp 
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from parameters
[ RUN      ] parameters.test_parametesr
Parameter name: bar
Parameter name: foo
Parameter name: foo.first
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:57: Failure
Value of: name == "foo" || name == "bar"
  Actual: false
Expected: true
Parameter name: foo.second
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:57: Failure
Value of: name == "foo" || name == "bar"
  Actual: false
Expected: true
footruebool
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:69: Failure
Value of: parameter.to_string().c_str()
  Actual: "true"
Expected: "2"
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:70: Failure
Value of: parameter.get_type_name().c_str()
  Actual: "bool"
Expected: "int"
bar1.450000double
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:74: Failure
Value of: parameter.to_string().c_str()
  Actual: "1.450000"
Expected: "hello"
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:75: Failure
Value of: parameter.get_type_name().c_str()
  Actual: "double"
Expected: "string"
baz2integer
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:79: Failure
Value of: parameter.to_string().c_str()
  Actual: "2"
Expected: "1.45"
/home/tfoote/work/ros2/test_param/src/ros2/system_tests/test_communication/test/test_parameters.cpp:80: Failure
Value of: parameter.get_type_name().c_str()
  Actual: "integer"
Expected: "double"
tested parameters for 0.277822 seconds
[  FAILED  ] parameters.test_parametesr (3384 ms)
[----------] 1 test from parameters (3384 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (3384 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] parameters.test_parametesr
```

tfoote:
https://github.com/ros2/rmw_connext/issues/66 should fix the services issue

tfoote:
This is blocked by #29 before it can be merged.

test_rclcpp/CMakeLists.txt
```diff
@@ -131,6 +133,42 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    ament_add_gtest(local_parameters_cpp__${middleware_impl} "test/local_parameters.cpp")
+    add_dependencies(local_parameters_cpp__${middleware_impl} ${PROJECT_NAME})
```
dirk-thomas: Before using the gtest target you must check that it actually exists (e.g. when gtest is not found):

    if(TARGET local_parameters_cpp__${middleware_impl})

Same below.

esteve: I just added both to `CMakeLists.txt`, let me know if they look ok and I'll squash and merge this PR.

esteve:
I've removed the tests for the `to_json` functions since it's not clear if ros2/rclcpp#91 will be merged. I've pushed the tests to a separate branch https://github.com/ros2/system_tests/tree/test_json

esteve:
Just triggered the CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/284/
http://ci.ros2.org/job/ros2_batch_ci_windows/284/
http://ci.ros2.org/job/ros2_batch_ci_osx/210/

tfoote:
It looks like we still have intermittent failures in the services. 

The OSX build failed with: test_requester_replier_cpp__empty__rmw_opensplice_cpp.test_requester_replier
And the linux build failed with 
test_requester_replier_cpp__primitives__rmw_connext_cpp.test_requester_replier

I retriggered the linux and osx jobs: 
http://ci.ros2.org/job/ros2_batch_ci_linux/289/
http://ci.ros2.org/job/ros2_batch_ci_osx/214/

Windows still has 28 failures which is our baseline at the moment.

esteve:
@tfoote should we merge this PR? I don't see a reason why not, the intermittent failures are not introduced by this PR, and these changes will just make it easier to watch for errors in the parameters code.

test_rclcpp/CMakeLists.txt
```diff
@@ -135,6 +137,46 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    if(TARGET local_parameters_cpp__${middleware_impl})
+      ament_add_gtest(local_parameters_cpp__${middleware_impl} "test/local_parameters.cpp")
```
dirk-thomas: The cpp file contains a main function and should therefore not link against gtest's main function.

esteve: What's the solution here? Remove the main function from `local_parameters.cpp` or not add it with `ament_add_gtest` ?

tfoote: There's an argument to SKIP_LINKING_MAIN_LIBRARIES I'll take care of that and clean up the whitespace above before we merge.

tfoote:
I've amended for dirk's comment about the main. And cleared a few whitespace only changes. 

CI pending: 
http://ci.ros2.org/job/ros2_batch_ci_linux/293/
http://ci.ros2.org/job/ros2_batch_ci_osx/218/
http://ci.ros2.org/job/ros2_batch_ci_windows/292/

tfoote:
All tests pass except the one lower level test, tickted in #37  http://ci.ros2.org/job/ros2_batch_ci_osx/218/testReport/junit/%28root%29/test_requester_replier_cpp__primitives__rmw_opensplice_cpp/test_requester_replier/

PR #16, opened 2015-07-15
Add test for "has_bounded_size" message function

jacquelinekay:
Connects to https://github.com/ros2/rosidl/pull/29

Adds a test for message types and several new message types for the purposes of testing the `has_bounded_size()` function for messages.


dirk-thomas:
These test to not check anything related to communication and don't require an rmw implementation.
Sincethey only check the generated C++ code of the messages they should be added to the `rosidl_generator_cpp` package.

PR #20, opened 2015-07-29
add test_rclcpp package testing subscriptions and spinning for now

dirk-thomas:
The test currently fails with all rmw implementations.

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -1,0 +1,58 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(test_rclcpp)
+
+if(NOT WIN32)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+endif()
+
+find_package(ament_cmake REQUIRED)
+
+if(AMENT_ENABLE_TESTING)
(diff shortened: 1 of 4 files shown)


esteve:
+1

test_rclcpp/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+  if(NOT WIN32)
+    set(test_executable_extension "")
+    set(test_executable_subfolder "")
+  else()
+    set(test_executable_extension ".exe")
+    set(test_executable_subfolder "/${CMAKE_BUILD_TYPE}")
+    if("${test_executable_subfolder}" STREQUAL "/")
+      set(test_executable_subfolder "/Debug")
+    endif()
+  endif()
```
wjwwood: Where are these variables used?

dirk-thomas: Nowhere. I removed them. Thanks.

jacquelinekay:
Perhaps the timeout tests from ros2/system_tests/pull/19 should be in this package rather than `test_communication`. I can revisit that PR to put the test in `test_rclcpp` after this one gets merged if you agree.

dirk-thomas:
@jacquelinekay I agree the `test_rclcpp` is the better location for testing parameters.

PR #23, opened 2015-07-30
relax test to make OS X happy

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_osx/59/ passes all tests again.

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -91,17 +91,30 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
   ASSERT_EQ(1, counter);
 
   // while four messages have been published one callback should be triggered here
-  std::cout << "spin_node_once() - callback (2) expected" << std::endl;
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_subscription.cpp
```diff
@@ -91,17 +91,29 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
...
+  }
   ASSERT_EQ(3, counter);
 
   // check for all remaning calls
   std::cout << "spin_node_some() - callbacks (4 and 5) expected" << std::endl;
   executor.spin_node_some(node);
+  if (counter == 3 || counter == 4) {
+    // give the executor thread time to process the event
+    std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    std::cout << "spin_node_some() - callback (4 and 5) expected - trying again" << std::endl;
```
tfoote: or instead of and to match L112

dirk-thomas: I replaced the "4 and 5" with a condition string instead.

tfoote:
+1

PR #24, opened 2015-07-30
Added QoS settings

esteve:
Connects to ros2/ros2#66

### DIFF
--- a/test_rclcpp/test/test_timeout_subscriber.cpp
+++ b/test_rclcpp/test/test_timeout_subscriber.cpp
@@ -40,8 +40,12 @@ TEST(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscriber)
   auto start = std::chrono::steady_clock::now();
 
   auto node = rclcpp::Node::make_shared("test_timeout_subscriber");
+
+  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
+  custom_qos_profile.depth = 10;
+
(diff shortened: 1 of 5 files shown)


dirk-thomas:
+1

test_rclcpp/test/test_timeout_subscriber.cpp
```diff
@@ -40,8 +40,12 @@ TEST(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscriber)
   auto start = std::chrono::steady_clock::now();
 
   auto node = rclcpp::Node::make_shared("test_timeout_subscriber");
+
+  rmw_qos_policies_t custom_qos_policies = rmw_qos_policies_default;
```
jacquelinekay: I think `rmw_qos_policies_default` should be passed as a default argument to `rclcpp::create_subscription` and `rclcpp::create_publisher`. This could go in a future PR because it's another major API change that would touch a lot of code (since the callback argument would have to go before the qos_policies argument).

esteve: @jacquelinekay yeah, my first attempt did that. But ended up basically replacing the `queue_size` argument with `qos_policy` and didn't make it optional, to mimic the previous behavior that required a queue size to be passed. In any case, I think it's a good idea to have a default QoS policy set so users are not required to pass one every time.

jacquelinekay:
+1 (my suggested API change can happen later)

esteve:
@jacquelinekay I just created a ticket to address that ros2/ros2#80

PR #25, opened 2015-08-03
add test for timers

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_linux/133/
http://ci.ros2.org/job/ros2_batch_ci_osx/65/
http://ci.ros2.org/job/ros2_batch_ci_windows/92/

### DIFF
--- a/test_rclcpp/test/test_timer.cpp
+++ b/test_rclcpp/test/test_timer.cpp
@@ -1,0 +1,120 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <chrono>
+#include <iostream>
+
+#include <gtest/gtest.h>
+
+#include <rclcpp/rclcpp.hpp>
+
+TEST(test_time, timer_fire_regularly)
+{
+  rclcpp::init(0, nullptr);
+
+  auto node = rclcpp::Node::make_shared("test_timer_fire_regularly");
+
+  size_t counter = 0;
+  auto callback =
+    [&counter]() -> void
+    {
+      ++counter;
+      std::cout << "  callback() " << counter << std::endl;
+    };
+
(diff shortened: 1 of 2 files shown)


wjwwood:
You'll have to use spin or spin_once to test what I'm talking about. Spin some will result in a non-blocking spin, so timers could be handled when rmw_wait wakes up. You need a blocking call to rmw_wait or a timed call to rmw_wait in order to test what I was describing. That's why I suggested calling spin in a thread do it can block.

dirk-thomas:
I have added the test you describe in 51d1565988c26fc718b2d3de0e21ca2aa1497e4b.

test_rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+  auto shutdown =
+    []() -> void
+    {
+      rclcpp::shutdown();
+      std::cout << "shutdown()" << std::endl;
+    };
+
+  auto shutdown_timer = node->create_wall_timer(
+    std::chrono::duration_cast<std::chrono::nanoseconds>(4.5 * period), shutdown);
```
wjwwood: Using the mechanism you're testing in order to shutdown the test is not a reliable way to make sure the test stops. In fact if this test fails as I expect it to, this callback will not be executed, the test will hang, and the test runner will have to kill it.

test_rclcpp/test/test_timer.cpp
```diff
@@ -0,0 +1,126 @@
...
+  auto spinner =
+    [&executor, &node]() -> void
+    {
+      std::cout << "spin() until shutdown timer" << std::endl;
+      executor.add_node(node, false);
+      executor.spin();
+    };
+
+  auto thread = std::thread(spinner);
+  thread.join();
```
wjwwood: Why not just drop the shutdown timer, then do this:

```c++
auto thread = std::thread(spinner);
std::this_thread::sleep_for(period * 4);
rclcpp::shutdown();
thread.join();
```

dirk-thomas: Updated the test in 78fcfdba257c38ba5d8a537b8e0b1c1254bb2a70

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_linux/135/
http://ci.ros2.org/job/ros2_batch_ci_osx/67/
http://ci.ros2.org/job/ros2_batch_ci_windows/94/

wjwwood:
lgtm, thanks for iterating on it.

dirk-thomas:
With the increased timeout (9c2114d96ba6893dcf3b4180ea6d085fa6bd6c38) the tests pass:
http://ci.ros2.org/job/ros2_batch_ci_linux/136/
http://ci.ros2.org/job/ros2_batch_ci_osx/68/

I have also created ament/ament_cmake#19 to better handle this case.

PR #26, opened 2015-08-04
update subscription test to check correct deregistration

dirk-thomas:
The updated test checks that when the subscriber goes out of scope deregisters correctly.

The patch also replaces `std::cout` with `printf` - sorry for mixing it.

Currently the test fails since the subscriber does not deregister itself correctly. It seems like the subscription handle is not yet destroyed because something else hold a shared pointer of it.


dirk-thomas:
I have added the same test for timers which fails in the same way.

wjwwood:
+1

PR #27, opened 2015-08-11
add test to check for receiving callbacks for invalid data

dirk-thomas:
Connects to ros2/rmw_connext#70

### DIFF
--- a/test_communication/test/test_subscription_valid_data.cpp
+++ b/test_communication/test/test_subscription_valid_data.cpp
@@ -1,0 +1,75 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


test_communication/test/test_subscription_valid_data.cpp
```diff
@@ -0,0 +1,75 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <chrono>
+#include <iostream>
+#include <string>
+
+#include <rclcpp/rclcpp.hpp>
+
+#include <test_communication/msg/u_int32.hpp>
```
wjwwood: It's unrelated to this pr, but I'd say the transformation should not put an underscore between capital letters. `u_int32` is pretty silly looking, it should just be `uint32`. On the other hand, consider `RGBDImage` (obviously a manufactured example), which by my suggestion would be `rgbdimage`, should probably be `rgbd_image`. But even more wrong would be `r_g_b_d_image`. Obviously making the "correct" translation is not straight forward.

wjwwood:
+1

test_communication/test/test_subscription_valid_data.cpp
esteve: Agree, `u_int32` looks weird. +1 to not put an underscore between capital letters.

esteve:
+1

PR #28, opened 2015-08-18
adding tests for intra process communications

wjwwood:
Connects to ros2/ros2#59

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -89,6 +89,26 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    # test_intra_process
+    ament_add_gtest(
+      gtest_intra_process__${middleware_impl}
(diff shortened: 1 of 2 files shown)


jacquelinekay:
Are you going to test the case of multiple nodes with intra-process enabled here, or the case of 1 node with intra-process enabled and 1 with intra-process disabled?

wjwwood:
Yes....

Let's push those for round 2 where I actually implement subscription filtering because this test is horribly naive as is. So as most this test is testing that it just works in the best scenario.

I'll add a note on https://github.com/ros2/ros2/issues/91 so we don't forget: https://github.com/ros2/ros2/issues/91#issuecomment-132781068

jacquelinekay:
Sounds good.

wjwwood:
I should say this test is horribly naive as is because there is no subscription filtering. Anything more complex would likely break.

esteve:
+1

PR #30, opened 2015-08-19
add test with repeated publishers / subscribers

dirk-thomas:
The test fails for `rmw_connext_dynamic_cpp`:

http://ci.ros2.org/job/ros2_batch_ci_linux/228/
http://ci.ros2.org/job/ros2_batch_ci_osx/149/
http://ci.ros2.org/job/ros2_batch_ci_windows/231/

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -31,6 +31,26 @@ if(AMENT_ENABLE_TESTING)
     # get typesupport of rmw implementation to include / link against the corresponding interfaces
     get_rmw_typesupport(typesupport_impl "${middleware_impl}")
 
+    # test_repeated_publisher_subscriber
+    ament_add_gtest(
+      gtest_repeated_publisher_subscriber__${middleware_impl}
+      "test/test_repeated_publisher_subscriber.cpp"
+      TIMEOUT 15
+    )
+    if(TARGET gtest_repeated_publisher_subscriber__${middleware_impl})
+      target_link_libraries(gtest_repeated_publisher_subscriber__${middleware_impl}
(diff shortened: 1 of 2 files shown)


wjwwood:
lgtm

tfoote:
This needs a matching patch. We should move it to the backlog or ready?

Note one uncrustify failure too.

dirk-thomas:
The uncrustify error has already been fixed. I think this should stay in review until we either have a patch for rmw_connext_dynamic_cpp or a SSCCE which we can provide upstream (and temporary disable that test).

tfoote:
My interpretation of the in review column is that it's code which is ready to merge and needs review before merging. Once it has passed review it will be merged. 

To me this is more of an open bug report with attached unit test. The test itself can be reviewed, but there is still pending an implementation before it can be merged. Thus I would expect this to be filed in the backlog, or promoted to ready and in-progress as someone is addressing the underlying problem associated with the issue, just like if it was any other open ticket. I did the same thing with ros2/system_tests#29 it's a ticket with attached failing unit test. But it's not ready for merging thus it's not under review.

dirk-thomas:
I also disagree to move similar cases like ros2/system_tests#29 to the backlog but that is a topic for a separate offline discussion.

Back to this PR: I continued trying to reproduce it locally and wrote a standalone version of it. Seeing that this one worked I continued debugging `rmw_connext_dynamic_cpp` and came up with the following fix: ros2/rmw_connext#94

I triggered a new set of CI jobs:
http://ci.ros2.org/job/ros2_batch_ci_linux/248/
http://ci.ros2.org/job/ros2_batch_ci_osx/176/
http://ci.ros2.org/job/ros2_batch_ci_windows/253/

jacquelinekay:
+1

PR #31, opened 2015-08-20
fix timer test name

dirk-thomas:


jacquelinekay:
CI:

http://ci.ros2.org/job/ros2_batch_ci_linux/241 (pending)
http://ci.ros2.org/job/ros2_batch_ci_osx/161/
http://ci.ros2.org/job/ros2_batch_ci_windows/241/

dirk-thomas:
I cancelled the jobs since you will only really see the effect when the tests are failing which is not the case.

jacquelinekay:
I just started them to check that you didn't break anything :)

lgtm from my local testing

PR #32, opened 2015-08-20
Added test for QoS durability

esteve:
This PR adds a test for TRANSIENT_LOCAL durability.
connects to #214


esteve:
CI jobs:

http://ci.ros2.org/job/ros2_batch_ci_linux/432/
http://ci.ros2.org/job/ros2_batch_ci_osx/342/
http://ci.ros2.org/job/ros2_batch_ci_windows/484/

clalancette:
@esteve Sorry we never got a chance to look at this.  Since I was passing through this repository, I took the opportunity to rebase it, clean it up, and get all combinations of the tests working using the newer CMake infrastructure.  Hope you don't mind.  Have a look and let me know what you think.

mikaelarguedas:
Note: Do you think we could we extend the current publisher / subscriber and test_publisher_subscriber templates rather than adding another set of talker and listener scripts? This would also allow testing different message types and custom node names / topic names that have proven to cause flaky tests when nodes from the previous tests haven't finished shutting down properly

clalancette:
> Note: Do you think we could we extend the current publisher / subscriber and test_publisher_subscriber templates rather than adding another set of talker and listener scripts? This would also allow testing different message types and custom node names / topic names that have proven to cause flaky tests when nodes from the previous tests haven't finished shutting down properly

Hm, yeah, that's a good point.  It is probably possible; let me take a quick look and see how hard it will be.

esteve:
@clalancette  
> Hope you don't mind. Have a look and let me know what you think.

no problem at all! Thanks for looking into this and fleshing it out.

mikaelarguedas:
marking this as connecting to #214 and removing question label

jacobperron:
I'm going to close this due to its age and inactivity. We still have the open issue if someone wants to revisit testing QoS combinations: https://github.com/ros2/system_tests/issues/214

PR #34, opened 2015-08-25
update test to catch repeated service callbacks

dirk-thomas:
The requester will never send a request which has been answered a second time.

http://ci.ros2.org/job/ros2_batch_ci_linux/267/
http://ci.ros2.org/job/ros2_batch_ci_osx/187/
http://ci.ros2.org/job/ros2_batch_ci_windows/268/


tfoote:
+1

PR #35, opened 2015-08-25
update intra proc tests with different assumptions

wjwwood:
This just fixes the existing test.

The assumption that `spin_node_once` will run the callback you think it will does not hold with intra process.

### DIFF
--- a/test_rclcpp/test/test_intra_process.cpp
+++ b/test_rclcpp/test/test_intra_process.cpp
@@ -66,8 +66,18 @@ TEST(test_intra_process_within_one_node, nominal_usage) {
     ASSERT_EQ(0, counter);
 
     // wait for the first callback
-    printf("spin_node_once() - callback (1) expected\n");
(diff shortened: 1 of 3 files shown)


test_rclcpp/test/test_intra_process.cpp
```diff
@@ -67,7 +67,7 @@ TEST(test_intra_process_within_one_node, nominal_usage) {
 
     // wait for the first callback
     printf("spin_node_once() - callback (1) expected\n");
-    executor.spin_node_once(node);
+    executor.spin_node_some(node);
```
dirk-thomas: I think this is a race condition now. Before `spin_node_once` was blocking until the callback was triggered. Now with `spin_node_some` the callback might not yet be ready.

wjwwood: That's right, I made potential fix in 125aab3

wjwwood:
Connects to ros2/ros2#91

PR #38, opened 2015-08-27
use message_info.from_intra_process in test

wjwwood:
Connects to ros2/demos#12

### DIFF
--- a/test_rclcpp/test/test_intra_process.cpp
+++ b/test_rclcpp/test/test_intra_process.cpp
@@ -35,11 +35,15 @@ TEST(test_intra_process_within_one_node, nominal_usage) {
 
   size_t counter = 0;
   auto callback =
-    [&counter](const test_rclcpp::msg::UInt32::SharedPtr msg) -> void
+    [&counter](
+    const test_rclcpp::msg::UInt32::SharedPtr msg,
+    const rmw_message_info_t & message_info
+    ) -> void
     {
       ++counter;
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/test_intra_process.cpp
```diff
@@ -68,8 +72,8 @@ TEST(test_intra_process_within_one_node, nominal_usage) {
     // wait for the first callback
     {
       size_t i = 0;
-      while (counter == 0 && i < 10) {
-        printf("spin_node_once() - callback (1) expected - try %zu/10\n", i++);
+      while (counter == 0 && i < 2) {
```
jacquelinekay: Did you reduce the iterations here because you expect better performance from intra-process?

wjwwood: Because I do not think it should take more than that. This condition may not hold in the future, but 2 seemed just as arbitrary as 10 but is stricter.

jacquelinekay:
+1

test_rclcpp/test/test_intra_process.cpp
tfoote: If this counter might be tuned in the future it would make sense to use a variable or constant instead of hard coding it in 5 places.

tfoote:
+1 with a recommendation to use a constant for the loop bounds.

esteve:
+1

PR #45, opened 2015-09-15
Test read condition

esteve:
This PR adds an acceptance test for repeated requests to a service.

Connects to ros2/rclcpp#52

Depends on #42

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -19,8 +19,14 @@ if(AMENT_ENABLE_TESTING)
   set(message_files
     "msg/UInt32.msg"
   )
+
+  set(service_files
+    "srv/AddTwoInts.srv"
+  )
+
   rosidl_generate_interfaces(${PROJECT_NAME}
     ${message_files}
+    ${service_files}
     SKIP_INSTALL
   )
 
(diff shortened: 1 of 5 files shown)


test_rclcpp/CMakeLists.txt
```diff
@@ -14,13 +14,20 @@ find_package(ament_cmake REQUIRED)
 
 if(AMENT_ENABLE_TESTING)
   find_package(ament_lint_auto REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
```
dirk-thomas: This line is not necessary since the next line does that already.

test_rclcpp/CMakeLists.txt
```diff
@@ -175,6 +182,38 @@ if(AMENT_ENABLE_TESTING)
...
+      "${middleware_impl}"
+      "rclcpp")
+    add_executable(test_services_server_cpp__${middleware_impl} "test/test_services_server.cpp")
+    add_dependencies(test_services_server_cpp__${middleware_impl} ${PROJECT_NAME})
+    rosidl_target_interfaces(test_services_server_cpp__${middleware_impl}
+      ${PROJECT_NAME} ${typesupport_impl})
+    ament_target_dependencies(test_services_server_cpp__${middleware_impl}
+      "${middleware_impl}"
+      "rclcpp")
+    get_target_property(TEST_SERVICES_SERVER_EXECUTABLE test_services_server_cpp__${middleware_impl} LOCATION)
```
dirk-thomas: This should do the same as ament/ament_cmake#29

Same below.

test_rclcpp/CMakeLists.txt
esteve: Fixed in the `test-services` branch.

test_rclcpp/CMakeLists.txt
esteve: Fixed in the `test-services` branch.

test_rclcpp/test/test_services_client.cpp
```diff
@@ -0,0 +1,82 @@
...
+    executor, node, result1);
+
+  EXPECT_EQ(15, result1.get()->sum);
+
+  EXPECT_EQ(19, result2.get()->sum);
+}
+
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
```
dirk-thomas: Pass `0, nullptr` instead of `arc, argv`.

Same below.

dirk-thomas:
I would propose to close this in favor of ros2/system_tests#57.

PR #44, opened 2015-09-04
Enable parameters tests

esteve:
Make sure that parameters tests are built and run.

PR #46, opened 2015-09-15
Added GTest include dir

esteve:
Spun off #44, adds GTest include dir. This change however doesn't take effect because the related test is not enabled.


dirk-thomas:
+1

PR #47, opened 2015-09-15
Call rclcpp::init only once

esteve:
This PR moves calls to `rclcpp::init` into a main function to ensure that it's only being called once.

### DIFF
--- a/test_rclcpp/test/test_timer.cpp
+++ b/test_rclcpp/test/test_timer.cpp
@@ -27,8 +27,6 @@
 #endif
 
 TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_fire_regularly) {
-  rclcpp::init(0, nullptr);
-
   auto node = rclcpp::Node::make_shared("test_timer_fire_regularly");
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/test_timer.cpp
```diff
@@ -134,3 +130,11 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_during_wait) {
   std::chrono::duration<float> diff = (end - start);
   printf("running for %.3f seconds\n", diff.count());
 }
+
+int main(int argc, char ** argv)
+{
+  // NOTE: use custom main to ensure that rclcpp::init is called only once
+  rclcpp::init(argc, argv);
```
dirk-thomas: Does this still work when passing GTest specific command line arguments? Shouldn't  it continue to pass the arguments `0, nullptr`?

esteve: Fixed.

dirk-thomas:
LGTM.

PR #49, opened 2015-09-28
tests for new parameter to_string API

tfoote:
connects to ros2/rclcpp#91

### DIFF
--- a/test_rclcpp/test/local_parameters.cpp
+++ b/test_rclcpp/test/local_parameters.cpp
@@ -21,6 +21,35 @@
 
 #include "parameter_fixtures.hpp"
 
+TEST(parameters, to_string) {
+  rclcpp::parameter::ParameterVariant pv("foo", "bar");
+  rclcpp::parameter::ParameterVariant pv2("foo2", "bar2");
+  std::string json_dict = rclcpp::parameter::to_json(pv);
+  EXPECT_STREQ(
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/local_parameters.cpp
```diff
@@ -21,6 +21,33 @@
 
 #include "parameter_fixtures.hpp"
 
+TEST(parameters, to_string) {
+  rclcpp::parameter::ParameterVariant pv("foo", "bar");
+  rclcpp::parameter::ParameterVariant pv2("foo2", "bar2");
+  std::string json_dict = rclcpp::parameter::to_json(pv);
+  EXPECT_STREQ(std::string(
+      "{\"name\": \"foo\", \"type\": \"string\", \"value\": \"bar\"}").c_str(), json_dict.c_str());
```
dirk-thomas: Why is the first argument using this:

    std::string("...").c_str()

dirk-thomas:
This patch results in three new cpplint errors:

    .../test_rclcpp/test/local_parameters.cpp:40:  Lines should be <= 100 characters long  [whitespace/line_length] [2]
    .../test_rclcpp/test/local_parameters.cpp:44:  Should have a space between // and comment  [whitespace/comments] [4]
    .../test_rclcpp/test/local_parameters.cpp:35:  Add #include <vector> for vector<>  [build/include_what_you_use] [4]

PR #51, opened 2015-10-08
disregard duplicate requests

### LINKED ISSUE #50: Intermittently failing requester/replier test
The test_requester_replier test is flaky on master.

e.g. on a recent test branch, it passed on osx but not Linux:
http://ci.ros2.org/job/ros2_batch_ci_linux/403/testReport/junit/(root)/test_requester_replier_cpp__empty__rmw_connext_dynamic_cpp/test_requester_replier/
http://ci.ros2.org/job/ros2_batch_ci_osx/300/

wjwwood:
fixes #50

### DIFF
--- a/test_communication/test/test_replier.cpp
+++ b/test_communication/test/test_replier.cpp
@@ -26,15 +26,12 @@ void reply(
   rclcpp::Node::SharedPtr node,
   const std::string & service_type,
   const std::vector<
-    std::pair<typename T::Request::SharedPtr, typename T::Response::SharedPtr>> & expected_services,
-  std::vector<bool> & received_messages
+    std::pair<typename T::Request::SharedPtr, typename T::Response::SharedPtr>> & expected_services
 )
 {
-  received_messages.assign(expected_services.size(), false);
-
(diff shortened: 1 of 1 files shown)


wjwwood:
As I see it, we either have to take this patch to fix #50 or increase the timeout on the requester such that it never encounters this race condition.

dirk-thomas:
If we don't want to check for duplicate requests I would suggest to remove the `received_messages` completely.

wjwwood:
Ok, if we agree on the approach to fix the flaky test then I'll remove it all.

wjwwood:
CI:
- http://ci.ros2.org/job/ros2_batch_ci_linux/407/
- http://ci.ros2.org/job/ros2_batch_ci_osx/304/
- http://ci.ros2.org/job/ros2_batch_ci_windows/453/

dirk-thomas:
+1 (if CI passes as expected)

esteve:
+1

wjwwood:
The Windows job has test failures, but are unrelated in difference from the master branch, I ran a master branch job on Windows right after the one for this branch for comparison:

http://ci.ros2.org/job/ros2_batch_ci_windows/454/

PR #52, opened 2015-10-08
Reduce test times

dirk-thomas:
http://ci.ros2.org/job/ros2_batch_ci_linux/406/
http://ci.ros2.org/job/ros2_batch_ci_osx/303/
http://ci.ros2.org/job/ros2_batch_ci_windows/452/

### DIFF
--- a/test_communication/test/test_publisher_subscriber.cpp
+++ b/test_communication/test/test_publisher_subscriber.cpp
@@ -33,8 +33,8 @@ void publish(
   auto publisher = node->create_publisher<T>(
     std::string("test_message_") + message_type, custom_qos_profile);
 
-  rclcpp::WallRate time_between_cycles(1);
-  rclcpp::WallRate time_between_messages(10);
+  rclcpp::WallRate time_between_cycles(10);
+  rclcpp::WallRate time_between_messages(40);
   size_t cycle_index = 0;
   // publish all messages up to number_of_cycles times, longer sleep between each cycle
   while (rclcpp::ok() && cycle_index < number_of_cycles) {
--- a/test_communication/test/test_publisher_subscriber.cpp
+++ b/test_communication/test/test_publisher_subscriber.cpp
@@ -33,8 +33,8 @@ void publish(
   auto publisher = node->create_publisher<T>(
     std::string("test_message_") + message_type, custom_qos_profile);
 
-  rclcpp::WallRate time_between_cycles(10);
-  rclcpp::WallRate time_between_messages(40);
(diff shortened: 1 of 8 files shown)


test_rclcpp/test/test_timer.cpp
```diff
@@ -59,7 +59,7 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_fire_regularly) {
     printf("spin_node_some() for 4s\n");
     while (std::chrono::steady_clock::now() < start + 4.5 * period) {
       executor.spin_node_some(node);
-      std::this_thread::sleep_for(period / 25);
```
jacquelinekay: Hmm, won't dividing by a smaller number actually have a larger result?

dirk-thomas: Not really.

Before it was 1000 / 25 = 40ms.
Afterwards it is 100 / 10 = 10ms.

And usually sleeping less then 10ms doesn't make much sense anyway.

jacquelinekay: Got it, I didn't see that you reduced the period.

Then you may want to change the print message `spin_node_some() for 4s` in line 59 to `400ms`.

test_communication/test/test_publisher_subscriber.cpp
```diff
@@ -33,8 +33,8 @@ void publish(
   auto publisher = node->create_publisher<T>(
     std::string("test_message_") + message_type, custom_qos_profile);
 
-  rclcpp::WallRate time_between_cycles(1);
-  rclcpp::WallRate time_between_messages(10);
+  rclcpp::WallRate time_between_cycles(10);
```
jacquelinekay: While we're looking at this test, this variable has a misleading name. `time` implies it is some number of seconds, when actually it represents a rate.

test_rclcpp/test/test_timer.cpp
dirk-thomas: Good point. But who wants to maintain that string. Now it says "for some time" :wink:

test_communication/test/test_publisher_subscriber.cpp
dirk-thomas: Good point. Addressed in a separate commit: 5d2532c50fcdef807cc77e305f23b910e4bf32b2

jacquelinekay:
+1

wjwwood:
lgtm

esteve:
+1

PR #53, opened 2015-10-09
Test for shared_ptr<const T> callback type

jacquelinekay:
Connects to ros2/rclcpp#119

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -180,7 +180,55 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_c
 
   executor.spin_node_some(node);
   // spin up to 4 times with a 25 ms wait in between
-  for (size_t i = 0; i < 4 && counter == 0; ++i) {
+  for (unsigned long i = 0; i < 4 && counter == 0; ++i) {
+    printf("callback not called, sleeping and trying again\n");
+    std::this_thread::sleep_for(std::chrono::milliseconds(25));
+    executor.spin_node_some(node);
+  }
+  ASSERT_EQ(1, counter);
+}
+// Shortened version of the test for the ConstSharedPtr with info callback signature
+TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_const_with_info) {
+  auto node = rclcpp::Node::make_shared("test_subscription");
+
+  auto publisher = node->create_publisher<test_rclcpp::msg::UInt32>(
+    "test_subscription", rmw_qos_profile_default);
+
+  unsigned long counter = 0;
+  auto callback =
+    [&counter](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
+      const rmw_message_info_t & info) -> void
+    {
+      ++counter;
+      printf("  callback() %lu with message data %u\n", counter, msg->data);
+      ASSERT_EQ(counter, msg->data);
+      ASSERT_FALSE(info.from_intra_process);
+    };
+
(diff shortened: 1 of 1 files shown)


jacquelinekay:
http://ci.ros2.org/job/ros2_batch_ci_windows/460/

test_rclcpp/test/test_subscription.cpp
```diff
@@ -140,3 +140,50 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
...
+    executor.spin_node_some(node);
+    ASSERT_EQ(0, counter);
+
+    msg->data = 1;
+    publisher->publish(msg);
+    ASSERT_EQ(0, counter);
+
+    // wait for the first callback
+    printf("spin_node_once() - callback (1) expected\n");
+    executor.spin_node_once(node);
```
wjwwood: This could be flaky in the future. Relying on the next callback being the one you registered above is fragile. I'd wrap this in for loop to try multiple times or maybe use `spin_some`, but that might have its own issues.

jacquelinekay: oh right, I copied this before your fix went in

test_rclcpp/test/test_subscription.cpp
```diff
@@ -140,3 +140,50 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
...
 }
+
+// Shortened version of the test for the ConstSharedPtr callback signature
+TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_const) {
+  rclcpp::init(0, nullptr);
+
+  auto node = rclcpp::Node::make_shared("test_subscription");
+
+  rmw_qos_profile_t custom_qos_profile = rmw_qos_profile_default;
+  custom_qos_profile.depth = 10;
```
dirk-thomas: That seems to be the default now anyway? https://github.com/ros2/rmw/blob/65a952913a49b0f9c38088a8dc0504b6943ed609/rmw/include/rmw/qos_profiles.h#L39-L42

test_rclcpp/test/test_subscription.cpp
```diff
@@ -140,3 +140,50 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
   executor.spin_node_some(node);
   ASSERT_EQ(5, counter);
 }
+
+// Shortened version of the test for the ConstSharedPtr callback signature
+TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_const) {
+  rclcpp::init(0, nullptr);
```
dirk-thomas: Calling `rclcpp::init` in each test is not a good idea. Currently it doesn't do anything but instead it should only be called once for all tests.

test_rclcpp/test/test_subscription.cpp
```diff
@@ -140,3 +140,50 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
...
+      ++counter;
+      printf("  callback() %lu with message data %u\n", counter, msg->data);
+      ASSERT_EQ(counter, msg->data);
+    };
+
+  auto msg = std::make_shared<test_rclcpp::msg::UInt32>();
+  msg->data = 0;
+  rclcpp::executors::SingleThreadedExecutor executor;
+
+  {
```
dirk-thomas: This scope seems to be unnecessary.

jacquelinekay:
Last call before I merge this (the prerequisite rclcpp PR has been merged).

esteve:
+1

PR #56, opened 2015-10-14
add a test which uses the create_subscription with queue size api

wjwwood:
Connects to ros2/design#56.

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -238,51 +238,12 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_c
   ASSERT_EQ(1, counter);
 }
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Does this test have to replicate the whole testing with spin etc. or is it sufficient to only check the signature to create a subscription with a queue size?

wjwwood:
Well, I figure it was worth making sure a subscription created with this API would result in a functioning subscription, but I can slim the test down further if you don't think that's important.

dirk-thomas:
Related to my other comment (https://github.com/ros2/rclcpp/pull/128#issuecomment-148218292) I would think testing the functionality of the core functionality is enough. For the syntactic sugar API a simple test checking that a specific signature works would be enough.

wjwwood:
Ok I simplified the test to just creating a subscription. The tests passed locally, so I won't run another CI.

esteve:
+1

dirk-thomas:
lgtm

PR #57, opened 2015-10-15
add test with multiple service calls

dirk-thomas:
This new test implement exactly the same test for services as ros2/system_tests#20 does for pub/sub. It covers what has been tracked in ros2/rclcpp#52 for a while.

Without any patches the test fails since most rmw implementations currently don't wake up a second time: http://ci.ros2.org/job/ros2_batch_ci_osx/349/

With the two patches (ros2/rmw_connext#66 ros2/rmw_opensplice#60) the test passes since the rmw implementations check the read correctly (as suggested in https://github.com/ros2/rclcpp/issues/52#issuecomment-126523047): http://ci.ros2.org/job/ros2_batch_ci_osx/350/

Connect to ros2/rclcpp#52

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -133,6 +133,26 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    # test_multiple_service_calls
+    ament_add_gtest(
+      gtest_multiple_service_calls__${middleware_impl}
+      "test/test_multiple_service_calls.cpp"
+      TIMEOUT 15
+    )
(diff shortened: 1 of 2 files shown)


jacquelinekay:
Nice. Just for my own sanity, I'm going to build and the `fix_service_samples` branches locally and run the the multiple services test to make sure it's not flaky.

esteve:
+1

jacquelinekay:
The test passed 50 times on my machine. I feel good about that. +1

esteve:
I've merged the two PRs that fix the issue, this PR should be ready for merging.

PR #61, opened 2015-10-28
Replace unsigned long with uint32_t

esteve:
Use `uint32_t` instead of `unsigned long` to comply with `cpplint`

Connects to ament/ament_lint#34


tfoote:
+1

PR #65, opened 2015-11-05
Fix osx build

jacquelinekay:
Use enable_if to disable the `construct` method if the pointer type is const in the custom allocator.

### DIFF
--- a/test_rclcpp/test/test_allocator.cpp
+++ b/test_rclcpp/test/test_allocator.cpp
@@ -255,18 +255,19 @@ public:
 
   // construct and destroy should not be required to implement allocator_traits,
   // this is a bug in gcc 4.8 for usage in std::pair, std::map
-  template<typename _Up, typename ... _Args>
+  template<typename U, typename ... Args,
+  typename std::enable_if<!std::is_const<U>::value>::type * = nullptr>
   void
-  construct(_Up * __p, _Args && ... __args)
+  construct(U * ptr, Args && ... args)
   {
-    ::new(static_cast<void *>(__p))_Up(std::forward<_Args>(__args) ...);
+    ::new(ptr)U(std::forward<Args>(args) ...);
   }
 
-  template<typename _Up>
+  template<typename U>
   void
-  destroy(_Up * __p)
+  destroy(U * ptr)
   {
-    __p->~_Up();
+    ptr->~U();
   }
 
   // rebind should not be required to implement allocator_traits, this is a bug in gcc 4.8


test_rclcpp/test/test_allocator.cpp
```diff
@@ -255,18 +255,18 @@ struct InstrumentedAllocator : public allocator_pointer_traits<T>
...
   // this is a bug in gcc 4.8 for usage in std::pair, std::map
-  template<typename _Up, typename ... _Args>
-  void
-  construct(_Up * __p, _Args && ... __args)
+  template<typename U, typename ... Args>
+  typename std::enable_if<!std::is_const<U>::value>::type
+  construct(U * ptr, Args && ... args)
   {
-    ::new(static_cast<void *>(__p))_Up(std::forward<_Args>(__args) ...);
+    ::new(reinterpret_cast<void *>(ptr))U(std::forward<Args>(args) ...);
```
dirk-thomas: Why reinterpret instead of static here?

test_rclcpp/test/test_allocator.cpp
```diff
@@ -255,18 +255,18 @@ struct InstrumentedAllocator : public allocator_pointer_traits<T>
 
   // construct and destroy should not be required to implement allocator_traits,
   // this is a bug in gcc 4.8 for usage in std::pair, std::map
-  template<typename _Up, typename ... _Args>
-  void
-  construct(_Up * __p, _Args && ... __args)
+  template<typename U, typename ... Args>
+  typename std::enable_if<!std::is_const<U>::value>::type
```
esteve: Just a style comment, you can add the `enable_if` as a template argument so that the function signature stays the same, e.g.:

```c++
template<typename _Up, typename ... _Args, typename std::enable_if<!std::is_const<U>::value>::type * = nullptr>
void
construct(_Up * __p, _Args && ... __args)
```

test_rclcpp/test/test_allocator.cpp
jacquelinekay: That was leftover from experimenting with what would make it compile. Actually, no cast is needed because it will implicitly cast from the U pointer type to `void*`.

esteve:
+1

jacquelinekay:
the first run built on OSX and Windows, I'm rerunning Jenkins again after making a few tweaks based on comments

http://ci.ros2.org/job/ros2_batch_ci_linux/553/
http://ci.ros2.org/job/ros2_batch_ci_osx/421/
http://ci.ros2.org/job/ros2_batch_ci_windows/568/

PR #66, opened 2015-11-05
move add_node out of spin, sleep for exactly 4 periods

jacquelinekay:
I've launched Jenkins jobs that will repeat only the timer test until failure or until it's run 20 times:

http://ci.ros2.org/job/ros2_batch_ci_osx/425/
http://ci.ros2.org/job/ros2_batch_ci_windows/572

Update: the jobs passed for osx and Windows, I'm running it for Linux after running it locally:
http://ci.ros2.org/job/ros2_batch_ci_linux/555/

### DIFF
--- a/test_rclcpp/test/test_timer.cpp
+++ b/test_rclcpp/test/test_timer.cpp
@@ -88,7 +88,7 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_during_wait) {
     [&counter]() -> void
     {
       ++counter;
-      printf("  callback() %4u", counter);
+      printf("  callback() %4u\n", counter);
     };
 
   rclcpp::executors::SingleThreadedExecutor executor;
@@ -96,6 +96,7 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_during_wait) {
   std::chrono::milliseconds period(100);
   auto timer = node->create_wall_timer(
     std::chrono::duration_cast<std::chrono::nanoseconds>(period), callback);
+  executor.add_node(node, false);
 
   // start condition
   ASSERT_EQ(0, counter);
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_timer.cpp
```diff
@@ -108,16 +109,15 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_during_wait) {
...
       printf("spin() until shutdown\n");
-      executor.add_node(node, false);
       executor.spin();
     };
 
   auto thread = std::thread(spinner);
-  printf("sleeping for 4 periods\n");
-  std::this_thread::sleep_for(4 * period);
+  printf("sleeping for 3.5 more periods\n");
+  std::this_thread::sleep_for(std::chrono::duration_cast<std::chrono::nanoseconds>(3.5 * period));
```
dirk-thomas: Why 3.5? In combination with the 0.5 sleep earlier that make it align exactly with a timer event and therefore very likely undeterministic.

jacquelinekay: I think one of the "problems" here is that the original test may not accurately reflect the current behavior of timers:

"Subtracting the loop rate period ensures that the callback gets triggered on the first call to check_and_trigger."

from https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/timer.hpp#L113

If t=0 is the first iteration of "spin", i is the iteration, and p is the timer period, we should expect that the callback triggers at t=0, t=p, t=2p... t=i*p.

Previously the test went like this:

1. Create the timer
2. Sleep for 0.5 period (this is BEFORE t=0, because we haven't spun yet)
3. In thread 1, add the node to the executor and call spin. In thread 2, sleep for 4 periods.

we stop spinning at t=4p, which is right at the instant where callback 5 should fire (the test expectation is to only see up to callback 4). Because the timer is inherently a little late, we see that callback 5 doesn't fire most of the time, but sometimes the Windows and OSX thread schedulers are early.

The tests that I launched pass on Windows and OSX (Linux is still running) because they do the following:

1. Create the timer
2. Sleep for 0.5 period (again, before t=0, so this does nothing to increment the timer)
3. In thread one, spin. In thread 2, sleep for 3.5 periods, which is time 3.5p. Between t=3p and t=4p we should have expected to see 4 callbacks (because we started at 0). The extra 0.5 p gives us plenty of time to wait for the late callback to trigger.

You could argue that we need to refactor timers to trigger for the first time on t=p, not t=0, or make it a toggleable option.

dirk-thomas: Wow, I was not aware that this is the current behavior of a timer. We should definitely change it to fire at `now() + period` for the first time.

Also I don't see what the `spin()` call should have to do with it. When I create the timer at `t1` I expect the first callback at `t1 + period` independent on when `spin()` was invoked (obviously only when some executor is handling it, otherwise the callback gets delayed until it is being handled).

jacquelinekay:
ok, based on discussion here and offline we should close this one and consider this branch that changes the behavior of timers instead:

ros2/rclcpp#146

PR #68, opened 2015-11-06
fix flaky shared_ptr_const test cases

jacquelinekay:
The `shared_ptr_const*` test cases of `test_subscription` fail intermittently on OSX and almost always fail in the nightly jobs:

http://ci.ros2.org/view/nightly/job/ros2_batch_ci_osx_nightly/101/testReport/junit/%28root%29/test_subscription__rmw_connext_dynamic_cpp/subscription_shared_ptr_const_with_info/
http://ci.ros2.org/view/nightly/job/ros2_batch_ci_osx_nightly/100/testReport/%28root%29/test_subscription__rmw_connext_cpp/subscription_shared_ptr_const_with_info/
http://ci.ros2.org/view/nightly/job/ros2_batch_ci_osx_nightly/99/testReport/junit/%28root%29/test_subscription__rmw_connext_dynamic_cpp/subscription_shared_ptr_const/

I take responsibility for this as the author of these tests--I think the problem is that I don't sleep after intitializing rclcpp entities and before spinning.

I also changed the subscription_and_spinning test to use `spin_once` with a timeout in several cases, instead of blocking and then calling spin_once without a timeout. This is more likely to give successful results and may be marginally faster, because spin_once will block asynchronously until the middleware is ready.

I started a job using `repeat-until-fail 20` and a regex to run only `test_subscription`:

http://ci.ros2.org/job/ros2_batch_ci_osx/440/

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -74,6 +74,11 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
     // wait for the first callback
     printf("spin_node_once() - callback (1) expected\n");
     executor.spin_node_once(node);
+    size_t i = 0;
+    while (counter < 1 && i < 2) {
+      executor.spin_node_once(node, std::chrono::milliseconds(25));
+      printf("spin_node_once(blocking for 25 ms) - callback (2) expected - try %zu/2\n", ++i);
+    }
     ASSERT_EQ(1, counter);
 
     // nothing should be pending here
@@ -99,9 +104,8 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
     executor.spin_node_once(node, std::chrono::milliseconds(0));
     if (counter == 1) {
       // give the executor thread time to process the event
-      std::this_thread::sleep_for(std::chrono::milliseconds(25));
-      printf("spin_node_once(nonblocking) - callback (2) expected - trying again\n");
-      executor.spin_node_once(node, std::chrono::milliseconds(0));
+      printf("spin_node_once(blocking for 25 ms) - callback (2) expected - trying again\n");
+      executor.spin_node_once(node, std::chrono::milliseconds(25));
     }
     ASSERT_EQ(2, counter);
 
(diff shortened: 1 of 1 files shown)


jacquelinekay:
Update: the shared_ptr test cases succeeded in that CI job, but the only case in the test that previously worked now fails (subscription_and_spinning). I think that may be due to the new behavior of timers. Looking into it.

jacquelinekay:
New repeated builds:

http://ci.ros2.org/job/ros2_batch_ci_osx/444/
http://ci.ros2.org/job/ros2_batch_ci_windows/595

jacquelinekay:
I made a few changes based on feedback from @wjwwood so that we don't always block for 25 ms in each case, as a small speed optimization. I've folded in almost all of the sleeps into `spin_once` with a timeout. These calls can block up to 25 ms or however long it takes for work to come in from DDS.

http://ci.ros2.org/job/ros2_batch_ci_osx/446/
http://ci.ros2.org/job/ros2_batch_ci_windows/602/

wjwwood:
lgtm, thanks for looking into it.

jacquelinekay:
argh. New OSX test failed.

jacquelinekay:
I did some local testing and it appears that putting `printf` before the first call to `spin_some` fixes the test. Rerunning Jenkins again to confirm.

http://ci.ros2.org/job/ros2_batch_ci_osx/454/

I actually don't think the pattern of waiting and checking with spin_some works; what matters is that the system blocks/context switches when `printf` is called. I experimented with calling `std::this_thread::yield` to try to force a context switch right after the publisher and subscriber objects are created, and it had similar results to adding a printf in the same spot (i.e. the test no longer seems flaky).

I won't pretend to understand everything that's going on here. I suspect that a context switch from the main thread after initializing all Connext datareaders/writers and before calling rmw_wait is a key component of more deterministic startup behavior. Hopefully the Jenkins build will show that this branch makes the test deterministic enough to fix the nightly build.

dirk-thomas:
Well, but we can't recommend user land code to call `printf()` to make stuff work. So we have to come up with a different fix.

gerkey:
This is definitely something that we need to resolve for alpha3, but it can wait until some other work on allocators is done.  In general, we need to make connext startup deterministic (without timed sleeps or printfs).

jacquelinekay:
I've ticketed ros2/rmw_connext#112 to cover the nondeterministic startup behavior in Connext and I'm hunting for duplicate tickets now.

jacquelinekay:
I haven't seen these test cases fail on the nightly builds for a while now--the problem seems to be services. Therefore I am closing this PR.

PR #71, opened 2015-11-11
Add new case to multiple_service_calls for "n" clients (currently 5)

jacquelinekay:
I made a similar test to this in my upcoming multithreaded test and I wanted to make a single-threaded version of it as a sanity check. This test case passes, but the multi-threaded case does not.

update: I got the multi-threaded version of this test case working in ros2/system_tests#72, but I'll keep this PR in case the multi-threaded executor breaks in the future.

Depends on ros2/rclcpp#153 and ros2/rmw_opensplice#95

### DIFF
--- a/test_rclcpp/test/test_multiple_service_calls.cpp
+++ b/test_rclcpp/test/test_multiple_service_calls.cpp
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 #include <iostream>
+#include <utility>
+#include <vector>
 
 #include "gtest/gtest.h"
 
@@ -34,15 +36,15 @@ void handle_add_two_ints(
   response->sum = request->a + request->b;
 }
 
-TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_service_calls) {
+TEST(CLASSNAME(test_two_service_calls, RMW_IMPLEMENTATION), two_service_calls) {
   rclcpp::init(0, nullptr);
 
-  auto node = rclcpp::Node::make_shared("test_multiple_service_calls");
+  auto node = rclcpp::Node::make_shared("test_two_service_calls");
 
   node->create_service<test_rclcpp::srv::AddTwoInts>(
-    "test_multiple_service_calls", handle_add_two_ints);
+    "test_two_service_calls", handle_add_two_ints);
 
-  auto client = node->create_client<test_rclcpp::srv::AddTwoInts>("test_multiple_service_calls");
+  auto client = node->create_client<test_rclcpp::srv::AddTwoInts>("test_two_service_calls");
 
   auto request1 = std::make_shared<test_rclcpp::srv::AddTwoInts::Request>();
   request1->a = 1;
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -34,7 +34,7 @@ void handle_add_two_ints(
   response->sum = request->a + request->b;
 }
 
-TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_service_calls) {
+TEST(CLASSNAME(test_two_service_calls, RMW_IMPLEMENTATION), two_service_calls) {
```
dirk-thomas: Please also update the node and topic names accordingly. Otherwise they collide with the new test.

jacquelinekay:
This test fails on Jenkins on Opensplice only, I think it may be related to ros2/rmw_opensplice#47

test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -70,3 +70,56 @@ TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_servic
...
+  rclcpp::init(0, nullptr);
+
+  auto node = rclcpp::Node::make_shared("test_multiple_service_calls");
+
+  node->create_service<test_rclcpp::srv::AddTwoInts>(
+    "test_multiple_service_calls", handle_add_two_ints);
+
+  using ClientRequestPair =
+      std::pair<rclcpp::client::Client<test_rclcpp::srv::AddTwoInts>::SharedPtr,
+      test_rclcpp::srv::AddTwoInts::Request::SharedPtr>;
```
dirk-thomas: I find these two lines hard to read. What about the following which makes the pair more visible:

    using ClientRequestPair = std::pair<
        rclcpp::client::Client<test_rclcpp::srv::AddTwoInts>::SharedPtr,
        test_rclcpp::srv::AddTwoInts::Request::SharedPtr>;

test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -69,4 +71,60 @@ TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_servic
...
+  }
+
+  std::vector<SharedFuture> results;
+  // Send all the requests
+  for (auto & pair : client_request_pairs) {
+    results.push_back(pair.first->async_send_request(pair.second));
+  }
+
+  printf("Sending %u requests...\n", n);
+  fflush(stdout);
```
dirk-thomas: Can this be moved above the lines which actually send the requests?

test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -69,4 +71,60 @@ TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_servic
...
+        rclcpp::shutdown();
+      }
+    };
+  auto timer = node->create_wall_timer(std::chrono::nanoseconds(3000000), timer_callback);
+
+  rclcpp::spin(node);
+
+  // Check the status of all futures
+  for (uint32_t i = 0; i < results.size(); i++) {
+    printf("Got request #%u\n", i);
```
dirk-thomas: The result has not been received yet. This print should be moved one line down.

Also flush afterwards?

test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -69,4 +71,64 @@ TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_servic
   rclcpp::spin_until_future_complete(node, result2);
   printf("Received second reply\n");
   EXPECT_EQ(2, result2.get()->sum);
+  // The first result should still be 1.
+  EXPECT_EQ(1, result1.get()->sum);
+}
+
+TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_clients) {
+  uint32_t n = 5;
```
dirk-thomas: Could this be `const`?

test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -69,4 +71,64 @@ TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_servic
...
+
+  executor.add_node(node);
+
+  executor.spin();
+
+  // Check the status of all futures
+  for (uint32_t i = 0; i < results.size(); i++) {
+    ASSERT_EQ(std::future_status::ready, results[i].wait_for(std::chrono::seconds(0)));
+    EXPECT_EQ(results[i].get()->sum, 2 * i + 1);
+    printf("Got request #%u\n", i);
```
dirk-thomas: The output should be `Got response#`, right?

The printing could also be moved one line up. Then it prints the message after it received the response and after that it validates its content.

test_rclcpp/test/test_multiple_service_calls.cpp
```diff
@@ -69,4 +71,64 @@ TEST(CLASSNAME(test_multiple_service_calls, RMW_IMPLEMENTATION), multiple_servic
...
+      }
+    };
+  auto timer = node->create_wall_timer(std::chrono::nanoseconds(3000000), timer_callback);
+
+  executor.add_node(node);
+
+  executor.spin();
+
+  // Check the status of all futures
+  for (uint32_t i = 0; i < results.size(); i++) {
```
dirk-thomas: `++i`

dirk-thomas:
Beside the last comments: lgtm

jacquelinekay:
All comments are addressed, however, this PR needs ros2/rclcpp#153 and ros2/rmw_opensplice#95 before it can be merged.

dirk-thomas:
Compiler warning: http://ci.ros2.org/job/ci_linux/625/warnings17Result/

dirk-thomas:
Compiler warning: http://ci.ros2.org/job/ci_osx/521/warnings7Result/

PR #72, opened 2015-11-12
Test for multithreaded execution

jacquelinekay:
This PR tests MultiThreadedExecutor `spin` and `spin_some` in 4 test cases:

1. Create 1 publisher and 2*N subscribers where N is the hardware concurrency of the machine, in an attempt to saturate the execution with work that must be done.
2. Same as above, but with intraprocess turned on.
3. Create 1 service and 2*N clients.
4. Create 1 publisher, 1 subscriber, and N timers that all attempt to publish a message on the publisher at the same time.

Connects to ros2/ros2#92

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -254,6 +254,26 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    # test_multithreaded
+    ament_add_gtest(
+      gtest_multithreaded__${middleware_impl}
+      "test/test_multithreaded.cpp"
+      TIMEOUT 15
+    )
+    if(TARGET gtest_multithreaded__${middleware_impl})
+      target_link_libraries(gtest_multithreaded__${middleware_impl}
+        ${_AMENT_EXPORT_ABSOLUTE_LIBRARIES}
+        ${_AMENT_EXPORT_LIBRARY_TARGETS})
+      target_compile_definitions(gtest_multithreaded__${middleware_impl}
+        PUBLIC "RMW_IMPLEMENTATION=${middleware_impl}")
+      add_dependencies(gtest_multithreaded__${middleware_impl} ${PROJECT_NAME})
+      rosidl_target_interfaces(gtest_multithreaded__${middleware_impl}
+        ${PROJECT_NAME} ${typesupport_impl})
+      ament_target_dependencies(gtest_multithreaded__${middleware_impl}
+        "${middleware_impl}"
+        "rclcpp")
+    endif()
+
     # test_local_parameters
     ament_add_gtest(
       gtest_local_parameters__${middleware_impl}
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -1,0 +1,289 @@
+// Copyright 2015 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0 //
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "gtest/gtest.h"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/executors.hpp"
+
+#include "test_rclcpp/msg/u_int32.hpp"
+#include "test_rclcpp/srv/add_two_ints.hpp"
+
+#ifdef RMW_IMPLEMENTATION
+# define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
+# define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+#else
+# define CLASSNAME(NAME, SUFFIX) NAME
+#endif
+
+static inline void multi_consumer_pub_sub_test(bool intra_process)
+{
+  std::string node_topic_name = "multi_consumer";
+  if (intra_process) {
+    node_topic_name += "_intra_process";
+  }
+
+  auto node = rclcpp::Node::make_shared(node_topic_name, intra_process);
+  auto callback_group = node->create_callback_group(
+    rclcpp::callback_group::CallbackGroupType::Reentrant);
+  auto pub = node->create_publisher<test_rclcpp::msg::UInt32>(node_topic_name, 16);
+
+  std::vector<rclcpp::Subscription<test_rclcpp::msg::UInt32>::SharedPtr> subscriptions;
+  std::atomic_uint counter;
+  counter = 0;
+
+  auto callback =
+    [&counter, &intra_process](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
+      const rmw_message_info_t & info) -> void
+    {
+      ++counter;
+      printf("callback() %4u with message data %u\n", counter.load(), msg->data);
+      ASSERT_EQ(intra_process, info.from_intra_process);
+    };
+
+  rclcpp::executors::MultiThreadedExecutor executor;
+  // Try to saturate the MultithreadedExecutor's thread pool with subscriptions
+  for (uint32_t i = 0; i < executor.get_number_of_threads(); i++) {
+    auto sub = node->create_subscription<test_rclcpp::msg::UInt32>(node_topic_name, 16, callback,
+        callback_group);
+    subscriptions.push_back(sub);
+  }
+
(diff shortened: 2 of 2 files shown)


test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+#ifdef RMW_IMPLEMENTATION
+# define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
+# define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+#else
+# define CLASSNAME(NAME, SUFFIX) NAME
+#endif
+
+static inline void multi_consumer_pub_sub_test(bool intra_process) {
+  rclcpp::init(0, nullptr);
```
dirk-thomas: This calls `init` twice which currently is not a problem might be as soon as the function does some real initialization work.

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -0,0 +1,197 @@
...
+  // Try to saturate the MultithreadedExecutor's thread pool with subscriptions
+  for (uint32_t i = 0; i < 2*executor.get_number_of_threads(); i++) {
+    auto sub = node->create_subscription<test_rclcpp::msg::UInt32>(node_topic_name, 10, callback);
+    subscriptions.push_back(sub);
+  }
+
+  executor.add_node(node);
+  auto msg = std::make_shared<test_rclcpp::msg::UInt32>();
+
+  // wait a moment for everything to initialize (TODO: fix nondeterministic startup behavior)
```
dirk-thomas: TODO with name (cpplint).

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -0,0 +1,197 @@
...
+  rclcpp::utilities::sleep_for(1_ms);
+
+  // sanity check that no callbacks have fired
+  executor.spin_once();
```
dirk-thomas: What event will be handled here? There should be any ready?

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -0,0 +1,197 @@
...
+  EXPECT_EQ(0, counter);
+
+  ++msg->data;
+  pub->publish(msg);
+
+  // test spin_some
+  // Expectation: The message was published and all subscriptions fired the callback.
+  executor.spin_some();
```
dirk-thomas: This will be a race condition since the published message might not yet have triggered an event.

test_rclcpp/test/test_multithreaded.cpp
jacquelinekay: But I don't have a custom main which calls init before starting the tests. Where does init get called twice?

Also, I call `shutdown` in all of the test cases, generally to stop the while loop in `spin`. Do I have to call `init` at the beginning of each test case since it is likely that the previous test case called `shutdown`?

test_rclcpp/test/test_multithreaded.cpp
jacquelinekay: The comment and the expectation here and below both suggest that this is verifying that no callbacks should fire before publishing a message. I can take this check out if you think it is redundant.

test_rclcpp/test/test_multithreaded.cpp
dirk-thomas: Both tests `multi_consumer_single_producer` and `multi_consumer_intraprocess` are calling this function and therefore `init`.

Currently we don't have a function symmetric to `init`. So I can't recommend anything but calling it only once in the main.

test_rclcpp/test/test_multithreaded.cpp
dirk-thomas: `spin_once` will block until there is one event to be dispatched. So I simply don't understand how this works.

test_rclcpp/test/test_multithreaded.cpp
jacquelinekay: Why can't `shutdown` be made symmetric to `init`?

dirk-thomas: It could be made symmetric. I was just referring to that it currently is not.

jacquelinekay: This (calling init, then shutdown, then init) actually is giving me a problem when I run all the test cases at once. I believe ros2/rclcpp#152 fixes the problem.

test_rclcpp/CMakeLists.txt
```diff
@@ -234,6 +234,26 @@ if(AMENT_ENABLE_TESTING)
         "rclcpp")
     endif()
 
+    # test_multithreaded
+    ament_add_gtest(
+      gtest_multithreaded__${middleware_impl}
+      "test/test_multithreaded.cpp"
+      TIMEOUT 30
```
dirk-thomas: How long is the expected runtime? Could the timeout be shorter?

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+  executor.add_node(node);
+  auto msg = std::make_shared<test_rclcpp::msg::UInt32>();
+
+  // wait a moment for everything to initialize
+  // TODO(jacquelinekay): fix nondeterministic startup behavior
+  rclcpp::utilities::sleep_for(1_ms);
+
+  // sanity check that no callbacks have fired
+  executor.spin_once();
```
dirk-thomas: This comment (https://github.com/ros2/system_tests/pull/72#discussion_r44692677) seems to still apply?

jacquelinekay:
I'm moving this back to In Progress to work on a segfault I've discovered in this test.

jacquelinekay:
With the fix @gerkey pushed yesterday, I ran this test on repeat and got a segfault in the multi_access_publisher_intra_process case after 245 iterations. Previously it would take 1-10 iterations to get a segfault on my machine. I also no longer see deadlock.

I will investigate the segfault and try to reproduce it. The core dump appears to have a corrupt stack and the code wasn't built in Debug mode.

dirk-thomas:
For better debugging I would also recommend to consider using a debug build of the rmw impl. you are using.

gerkey:
@jacquelinekay, are you testing with the changes in #165?

jacquelinekay:
With the changes made in ros2/rclcpp#165, I have run the test for over 400 iterations without seeing a segfault or deadlock.

jacquelinekay:
I've rebased this one branch `intra_process_lock` since that set of branches is required to make this test pass reliably.

gerkey:
+1 (with ros2/rclcpp#165 required for reliable execution)

This test helped us find some nasty races.  Good stuff.

test_rclcpp/CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 
+#set (CMAKE_CXX_COMPILER "clang++")
+#set (CMAKE_LINKER "llvm-ld")
+
```
dirk-thomas: This should be removed before merging.

dirk-thomas:
This PR seem to be the reason that the master build on Windows is failing now: http://ci.ros2.org/job/ci_windows_opensplice/733/

Why did the previous CI jobs pass? Did you change further stuff after that?

dirk-thomas:
With the PR ros2/rclcpp#167 the code builds at least. But some tests fail: http://ci.ros2.org/job/ci_windows_opensplice/734/testReport/

jacquelinekay:
Those test cases pass reliably on my Linux machine. I suspect the failures could be due to quirks in the Windows thread scheduler.

dirk-thomas:
While running these tests I also saw the following output. The test passed but the values for the first iterations look pretty wrong. Shouldn't this fail the test somehow?

```
[ RUN      ] test_multithreaded__rmw_connext_dynamic_cpp.multi_consumer_intra_process
RTI Data Distribution Service Evaluation License issued to OSRF (OSRF01) dthomas@osrfoundation.org For non-production use only.
Expires on 08-dec-2015 See www.rti.com for more information.
callback()    1 with message data 3223089
callback()    2 with message data 3223089
callback()    3 with message data 3223089
callback()    4 with message data 3223089
callback()    1 with message data 1
callback()    2 with message data 1
callback()    3 with message data 1
callback()    4 with message data 1
callback()    5 with message data 2
callback()    6 with message data 2
callback()    7 with message data 2
callback()    8 with message data 2
callback()    9 with message data 3
callback()   10 with message data 3
callback()   11 with message data 3
callback()   12 with message data 3
callback()   13 with message data 4
callback()   14 with message data 4
callback()   15 with message data 4
callback()   16 with message data 4
callback()   17 with message data 5
callback()   18 with message data 5
callback()   19 with message data 5
callback()   20 with message data 5
[       OK ] test_multithreaded__rmw_connext_dynamic_cpp.multi_consumer_intra_process (3875 ms)
```

jacquelinekay:
The tests actually never check the data, they only check the number of times the callback was called (store in atomic uint called counter).

I also never see that error on my machine. Which platform are you running on? And is it every time you run the test case or does it happen randomly?

dirk-thomas:
The snippet is from http://ci.ros2.org/job/ci_osx/541/consoleFull at time 33:57.272.

PR #75, opened 2015-11-19
convert timer callback signature to return bool

jacquelinekay:
Connects to ros2/rclcpp#159

todo: implement "fire n times" test in test_timer.cpp


esteve:
+1

PR #76, opened 2015-11-20
Finite timer

jacquelinekay:
Connects to ros2/rclcpp#159

### DIFF
--- a/test_rclcpp/test/test_timer.cpp
+++ b/test_rclcpp/test/test_timer.cpp
@@ -131,6 +131,40 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_during_wait) {
   printf("running for %.3f seconds\n", diff.count());
 }
 
+
+TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), finite_timer) {
+  auto node = rclcpp::Node::make_shared("finite_timer");
+
+  uint32_t counter = 0;
+  auto callback =
+    [&counter](rclcpp::timer::TimerBase & timer) -> void
+    {
+      ++counter;
+      printf("  callback() %4u\n", counter);
+      // After spinning 3 times, cancel the timer
+      if (counter == 2) {
+        timer.cancel();
+      }
+    };
+
+  // start condition
+  ASSERT_EQ(0, counter);
+
+  std::chrono::milliseconds period(10);
+
+  {
+    auto timer = node->create_wall_timer(
+      std::chrono::duration_cast<std::chrono::nanoseconds>(period), callback);
+    // spin a few times
+    for (uint32_t i = 0; i < 6; ++i) {
+      std::this_thread::sleep_for(period);
+      rclcpp::spin_some(node);
+    }
+
+    EXPECT_EQ(counter, 2);
+  }
+}
+
 int main(int argc, char ** argv)
 {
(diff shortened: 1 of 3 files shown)


test_rclcpp/test/test_executor.cpp
```diff
@@ -28,12 +28,15 @@
 TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), recursive_spin_call) {
   rclcpp::executors::SingleThreadedExecutor executor;
   auto node = rclcpp::Node::make_shared("recursive_spin_call");
-  auto timer = node->create_wall_timer(0_s, [&executor]() {
+  auto timer = node->create_wall_timer(0_s,
```
dirk-thomas: Please wrap either all arguments or none. In this case add a newline after the opening parenthesis.

test_rclcpp/test/test_executor.cpp
```diff
@@ -28,12 +28,15 @@
...
+      [&executor](rclcpp::timer::TimerBase & timer) {
     ASSERT_THROW(executor.spin_some(), std::runtime_error);
     ASSERT_THROW(executor.spin_once(), std::runtime_error);
     ASSERT_THROW(executor.spin(), std::runtime_error);
     executor.cancel();
-  });
+    timer.cancel();
+  }
+    );
```
dirk-thomas: Weird indentation.

test_rclcpp/test/test_executor.cpp
```diff
@@ -28,12 +28,15 @@
...
   auto node = rclcpp::Node::make_shared("recursive_spin_call");
-  auto timer = node->create_wall_timer(0_s, [&executor]() {
+  auto timer = node->create_wall_timer(0_s,
+      [&executor](rclcpp::timer::TimerBase & timer) {
     ASSERT_THROW(executor.spin_some(), std::runtime_error);
     ASSERT_THROW(executor.spin_once(), std::runtime_error);
     ASSERT_THROW(executor.spin(), std::runtime_error);
     executor.cancel();
-  });
+    timer.cancel();
```
dirk-thomas: Why does the timer need to be cancelled if the executor is cancelled?

test_rclcpp/test/test_timer.cpp
```diff
@@ -131,6 +131,40 @@ TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_during_wait) {
   printf("running for %.3f seconds\n", diff.count());
 }
 
+
+TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), finite_timer) {
+  auto node = rclcpp::Node::make_shared("test_timer_fire_regularly");
```
dirk-thomas: Please use a related name for the test and the node.

test_rclcpp/test/test_executor.cpp
jacquelinekay: Hmm, true, this is redundant.

There is a situation where cancelling the timer without cancelling the executor is useful: when the timer callback is publishing messages for a finite amount of time in-process, the timer can cancel itself after the last publish, wait some amount of time for subscriber to get the message, then cancel the executor...

In this test case, since there is no publish/subscriber, executor.cancel() is enough. Also, it never makes sense to call executor.cancel() before timer.cancel(), as I have done here and in the other test case.

esteve:
+1

PR #77, opened 2015-11-23
Improvements to Allocator test

jacquelinekay:
This PR attempts to reduce code duplication in the allocator test by using a test fixture.

It also adds "verbose" (default=false) and "ignore_middleware_tokens" (default=true) booleans to the test cases. Adding "verbose" or "all_tokens" to the invocation of the executable will toggle these options.

### DIFF
--- a/test_rclcpp/test/test_allocator.cpp
+++ b/test_rclcpp/test/test_allocator.cpp
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 #include <memory>
+#include <string>
+#include <vector>
 #include "gtest/gtest.h"
 
 #include "rclcpp/strategies/allocator_memory_strategy.hpp"
@@ -34,10 +36,21 @@
 
 #endif  // not WIN32
 
-static bool test_init = false;
+// TODO(jacquelinekay) improve this ignore rule (dogfooding or no allocations)
+static const size_t num_rmw_tokens = 5;
+static const char * rmw_tokens[num_rmw_tokens] = {"librmw", "dds", "DDS", "dcps", "DCPS"};
 
-/** Check a demangled stack backtrace of the caller function for the given tokens. */
-static inline bool check_stacktrace(const char ** tokens, size_t num_tokens, size_t max_frames = 6)
+static const size_t iterations = 1;
+
+static bool verbose = false;
+static bool ignore_middleware_tokens = true;
+static bool test_init = false;
+static bool fail = false;
+
+/** Check a demangled stack backtrace of the caller function for the given tokens.
+ ** Adapted from: https://panthema.net/2008/0901-stacktrace-demangled
+ **/
+inline bool check_stacktrace(const char ** tokens, size_t num_tokens, size_t max_frames = 6)
 {
 # ifndef WIN32
   bool match = false;
@@ -61,13 +74,13 @@ static inline bool check_stacktrace(const char ** tokens, size_t num_tokens, siz
   size_t funcnamesize = 256;
   char * funcname = static_cast<char *>(malloc(funcnamesize));
 
-  // fprintf(stderr, ">>>> stack trace:\n");
+  if (verbose) {
+    fprintf(stderr, ">>>> stack trace:\n");
+  }
+
   // iterate over the returned symbol lines. skip the first, it is the
   // address of this function.
   for (int i = 1; i < addrlen; i++) {
-    if (match) {
-      break;
-    }
     char * begin_name = 0, * begin_offset = 0, * end_offset = 0;
 
     // find parentheses and +address offset surrounding the mangled name:
@@ -90,10 +103,6 @@ static inline bool check_stacktrace(const char ** tokens, size_t num_tokens, siz
       *begin_offset++ = '\0';
       *end_offset = '\0';
 
-      // mangled name is now in [begin_name, begin_offset) and caller
-      // offset in [begin_offset, end_offset). now apply
-      // __cxa_demangle():
-
       int status;
       char * ret = abi::__cxa_demangle(begin_name,
           funcname, &funcnamesize, &status);
@@ -107,7 +116,9 @@ static inline bool check_stacktrace(const char ** tokens, size_t num_tokens, siz
             break;
           }
         }
-        // fprintf(stderr, "  %s : %s+%s\n", symbollist[i], funcname, begin_offset);
+        if (verbose) {
+          fprintf(stderr, "  %s : %s+%s\n", symbollist[i], funcname, begin_offset);
+        }
       } else {
         // demangling failed. Output function name as a C function with
         // no arguments.
@@ -119,7 +130,9 @@ static inline bool check_stacktrace(const char ** tokens, size_t num_tokens, siz
             break;
           }
         }
-        // fprintf(stderr, "  %s : %s()+%s\n", symbollist[i], begin_name, begin_offset);
+        if (verbose) {
+          fprintf(stderr, "  %s : %s()+%s\n", symbollist[i], begin_name, begin_offset);
+        }
       }
     } else {
       // couldn't parse the line? print the whole line.
@@ -129,14 +142,16 @@ static inline bool check_stacktrace(const char ** tokens, size_t num_tokens, siz
           break;
         }
       }
-      // fprintf(stderr, "  %s\n", symbollist[i]);
+      if (verbose) {
+        fprintf(stderr, "  %s\n", symbollist[i]);
+      }
     }
   }
 
   free(funcname);
   free(symbollist);
-  if (!match) {
-    fprintf(stderr, "WARNING: Called function did not match accepted tokens\n");
+  if (!ignore_middleware_tokens) {
+    return false;
   }
   return match;
 #else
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_allocator.cpp
```diff
@@ -34,9 +34,30 @@
 
 #endif  // not WIN32
 
+static const size_t num_rmw_tokens = 5;
+static const char * rmw_tokens[num_rmw_tokens] = {"librmw", "dds", "DDS", "dcps", "DCPS"};
```
dirk-thomas: This needs at least a TODO since it hardcodes values of unknown dependencies (not dogfooding).

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +431,50 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
-  global_runtime_deallocs = 0;
-  rclcpp::shutdown();
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  //argc and argv are modified by InitGoogleTest
+  if (argc > 0) {
+    std::string argv_str;
+    for (int i = 0; i < argc; ++i) {
```
dirk-thomas: `argv[0]` contains the called executable and shouldn't be considered for arguments.

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +431,50 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
+      argv_str += argv[i];
```
dirk-thomas: This logic is broken. Just concatenating all arguments without spaces and then search in this string is wrong. Consider passing `ver bos e` as three arguments.

esteve: You can also construct a vector from the array, so you don't have to build `argv_str` yourself:

```
std::vector<std::string> args(argv, argv + argc);
if (std::find(args.begin() + 1, args.end(), "verbose") != args.end()) {
```

test_rclcpp/test/test_allocator.cpp
jacquelinekay: Do you have any suggestions for how I can implement ignoring function calls from within the rmw implementation?

dirk-thomas: The rmw impl. packages would need to export that kind of information in order to be used here. That way other implemenations could do the same thing.

The same comment applies to a recent PR for launch: https://github.com/ros2/launch/pull/13#discussion_r44995670

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +432,47 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
-  global_runtime_deallocs = 0;
-  rclcpp::shutdown();
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  //argc and argv are modified by InitGoogleTest
+  if (argc > 0) {
+    std::vector<std::string> args(argv, argv + argc);
+    if (std::find(args.begin(), args.end(), "verbose") != args.end()) {
```
esteve: You can use `args.begin() + 1` to skip the first element of `argv`, so `std::find` will ignore the name of the program.

esteve: Or use `std::vector<std::string> args(argv + 1, argv + argc);` and `std::find` won't even have to ignore the name of the program.

esteve: I've pushed a commit that ignores the name of the program.

esteve:
+1

test_rclcpp/test/test_allocator.cpp
```diff
@@ -292,15 +304,91 @@ constexpr bool operator!=(const InstrumentedAllocator<T> &,
...
+  }
+
+  AllocatorTestFixture()
+  {
+    global_allocs = 0;
+    global_runtime_allocs = 0;
+    global_runtime_deallocs = 0;
+    global_deallocs = 0;
+    num_allocs = 0;
+    num_deallocs = 0;
```
dirk-thomas: These could be initialized with an initializer list:

    AllocatorTestFixture()
      : global_allocs(0),
        global_runtime_allocs(0),
        ...
    {

test_rclcpp/test/test_allocator.cpp
```diff
@@ -292,15 +304,91 @@ constexpr bool operator!=(const InstrumentedAllocator<T> &,
...
+
+  bool intra_process_;
+
+  size_t initialization_allocs_;
+  size_t initialization_deallocs_;
+
+  using UInt32Allocator = InstrumentedAllocator<test_rclcpp::msg::UInt32>;
+  using UInt32Deleter = rclcpp::allocator::Deleter<UInt32Allocator, test_rclcpp::msg::UInt32>;
+
+  void initialize(bool intra_process, std::string name)
```
dirk-thomas: Non-primitive types should be passed as `const &`.

test_rclcpp/test/test_allocator.cpp
dirk-thomas: Should this also set the values of the other member variables?

test_rclcpp/test/test_allocator.cpp
```diff
@@ -292,15 +304,91 @@ constexpr bool operator!=(const InstrumentedAllocator<T> &,
...
+      printf("Calls to global delete after spinning some: %zu\n", global_runtime_deallocs);
+      printf("Calls to allocator new after spinning some: %zu\n", runtime_allocs);
+      printf("Calls to allocator delete after spinning some: %zu\n", runtime_deallocs);
+    }
+    global_allocs = 0;
+    global_runtime_allocs = 0;
+    global_runtime_deallocs = 0;
+    global_deallocs = 0;
+    num_allocs = 0;
+    num_deallocs = 0;
```
dirk-thomas: Why set these variables to zero if the instance is destroyed?

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +432,47 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
+  //argc and argv are modified by InitGoogleTest
+  if (argc > 0) {
+    std::vector<std::string> args(argv + 1, argv + argc);
+    if (std::find(args.begin(), args.end(), "verbose") != args.end()) {
+      verbose = true;
+    }
+    if (std::find(args.begin(), args.end(), "all_tokens") != args.end()) {
+      ignore_middleware_tokens = false;
+    }
+  }
```
dirk-thomas: Should we prefix the two options with `--`? Maybe also print a short usage for `--help`?

esteve: Sounds good to me too, I'll give it a shot.

esteve: @jacquelinekay I've pushed the changes to a separate branch (https://github.com/ros2/system_tests/tree/printfs-help) so you can cherry pick the changes you prefer.

test_rclcpp/test/test_allocator.cpp
jacquelinekay: I believe an initializer list won't work here because these are global static variables.

They are static global variables because they are shared by the test fixture, all instances of the allocator class, and the static global member function "check_stacktrace".

However they get set and reset back to zero when each test case is initialized/deinitialized.

Alternatively, I could make these static member variables of the Allocator and declare a friend relationship between the Allocator and the test fixture.

dirk-thomas: Nvm I didn't see that these variables are actually globals.

If we ever try to run the tests in parallel this one looks like a problem.

I was also a bit confused by the class name `AllocatorTestFixture`. While it also contains a small *fixture* part it is mostly a subclass of `Test`. May be it shouldn't contain `Fixture` in its name or move that part into separate functions.

jacquelinekay: Some static global variables are going to be necessary because this test overrides global new and delete and modifies state that determines the fail condition for the test. Also, gtest EXPECT/ASSERT/FAIL macros cannot be called from within new and delete because they call new and delete. I can reduce the amount of global state we keep but they can't be eliminated entirely.

esteve:
I've rebased this branch and pushed it to `https://github.com/ros2/system_tests/tree/printfs-rebase`, I don't know if how I resolved the conflicts is how you'd have done it @jacquelinekay , feel free to pick the changes you deem appropriate.

jacquelinekay:
Rebased. If CI comes out clean I'll merge it.

http://ci.ros2.org/job/ci_linux/673/
http://ci.ros2.org/job/ci_osx/552/
http://ci.ros2.org/job/ci_windows_connext_dynamic/30/
http://ci.ros2.org/job/ci_windows_connext_static/29/
http://ci.ros2.org/job/ci_windows_opensplice/745

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +383,52 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
 
-  num_allocs = 0;
-  num_deallocs = 0;
-  global_allocs = 0;
-  global_deallocs = 0;
-  global_runtime_allocs = 0;
-  global_runtime_deallocs = 0;
-  rclcpp::shutdown();
+  if (argc > 0) {
```
dirk-thomas: This condition looks unnecessary.

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +383,52 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
-  printf("Calls to allocator new after spinning some: %zu\n", runtime_allocs);
-  printf("Calls to allocator delete after spinning some: %zu\n", runtime_deallocs);
+int main(int argc, char ** argv)
+{
+  rclcpp::init(argc, argv);
+  ::testing::InitGoogleTest(&argc, argv);
+  // argc and argv are modified by InitGoogleTest
+  std::vector<std::string> args(argv + 1, argv + argc);
+  bool help = (std::find(args.begin(), args.end(), "--help") != args.end());
+  if (help || argc == 0) {
```
dirk-thomas: When would `argc` ever be zero? And why should that trigger the help?

test_rclcpp/test/test_allocator.cpp
```diff
@@ -385,59 +383,52 @@ TEST(CLASSNAME(test_allocator, RMW_IMPLEMENTATION), unique_ptr) {
...
-  num_deallocs = 0;
-  global_allocs = 0;
-  global_deallocs = 0;
-  global_runtime_allocs = 0;
-  global_runtime_deallocs = 0;
-  rclcpp::shutdown();
+  if (argc > 0) {
+    if (std::find(args.begin(), args.end(), "--verbose") != args.end()) {
+      verbose = true;
+    }
```
dirk-thomas: This block could be simplified in one line like line 419 (`help`).

Same below.

test_rclcpp/test/test_allocator.cpp
jacquelinekay: I believe @esteve pushed these changes. I will take a quick look at it.

dirk-thomas: Since help is also not used anywhere else this could be boiled down to:

    if (std::find(args.begin(), args.end(), "--help") != args.end()) {
      print_help();

dirk-thomas:
These seem to fail on master: http://ci.ros2.org/job/ci_linux/679/testReport/

PR #82, opened 2015-12-04
Fix multithreaded test on Windows and Jenkins

jacquelinekay:
The test fails on Jenkins because it takes at ~20 seconds with Connext, but the timeout is set to 15.

The test also fails on Windows because of spurious wakeups in the scheduler. This PR adds a 1ms period busy wait to ensure that the condition is met.

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -44,8 +44,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
   auto pub = node->create_publisher<test_rclcpp::msg::UInt32>(node_topic_name, 16);
 
   std::vector<rclcpp::Subscription<test_rclcpp::msg::UInt32>::SharedPtr> subscriptions;
-  std::atomic_uint counter;
-  counter = 0;
+  std::atomic_uint counter(0);
 
   auto callback =
     [&counter, &intra_process](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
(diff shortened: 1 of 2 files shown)


jacquelinekay:
CI:

http://ci.ros2.org/job/ci_linux/665/
http://ci.ros2.org/job/ci_osx/547/
http://ci.ros2.org/job/ci_windows_connext_dynamic/26/
http://ci.ros2.org/job/ci_windows_connext_static/25/
http://ci.ros2.org/job/ci_windows_opensplice/739/

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -236,14 +238,21 @@ static inline void multi_access_publisher(bool intra_process)
   // use atomic
   std::atomic_uint timer_counter;
   timer_counter = 0;
+  std::atomic_uint subscription_counter;
+  subscription_counter = 0;
```
dirk-thomas: Can this be initialized on the same line? (Same two lines above)

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -236,14 +238,21 @@ static inline void multi_access_publisher(bool intra_process)
...
 
   const size_t iterations = 5 * executor.get_number_of_threads();
   auto timer_callback =
-    [&executor, &pub, &msg, &timer_counter, &iterations](rclcpp::timer::TimerBase & timer)
+    [&executor, &pub, &msg, &timer_counter, &subscription_counter,
+    &iterations](rclcpp::timer::TimerBase & timer)
```
dirk-thomas: Are the linters fine with wrapping it after the `]` and before the `(`?

dirk-thomas:
Other then the minor comments +1 to get the tests to pass.

jacquelinekay:
Argh. According to Jenkins, this PR causes the test to pass for all jobs except for the Windows Opensplice job, but I cannot reproduce this failure in my Windows VM (running Opensplice). I will merge this PR for the time being and revisit the failures later.

PR #84, opened 2015-12-07
Reverse ignore logic in allocator test

jacquelinekay:
ignore_middleware_tokens is the boolean in the allocator logic that causes the test to ignore allocations that originate in the DDS implementation.

--all-tokens is the command line argument flag that causes the test NOT to ignore middleware tokens.

This PR corrects the logic of the test to be consistent with this design.


dirk-thomas:
+1

PR #85, opened 2015-12-11
Added and increased waiting time in several rclcpp tests

richiprosima:
In order to solve http://ci.ros2.org/job/ci_linux/658/ we had to make some changes in several rclcpp tests. These changes are:
* test_allocator.cpp: Added new ```fastrtps``` token.
* test_multithreaded.cpp: Increased several sleeping times after publishing data, because executor is canceled before receiving all data.
* test_services_client.cpp: Added sleeping time after creating each client. It gives time to finish the discovery phase.
* test_subscriptions.cpp: Changed call ```spin_node_once``` for ```spin_node_some```. If sample 4 and 5 don't arrive in first try, in the second try only sample 4 is taken.

### DIFF
--- a/test_rclcpp/test/test_allocator.cpp
+++ b/test_rclcpp/test/test_allocator.cpp
@@ -37,8 +37,8 @@
 #endif  // not WIN32
 
 // TODO(jacquelinekay) improve this ignore rule (dogfooding or no allocations)
-static const size_t num_rmw_tokens = 5;
-static const char * rmw_tokens[num_rmw_tokens] = {"librmw", "dds", "DDS", "dcps", "DCPS"};
+static const size_t num_rmw_tokens = 6;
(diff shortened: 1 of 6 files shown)


jacquelinekay:
Hey @richiprosima:

with regard to the allocator test, I am aiming to phase out the ignore tokens soon, since it's a hack workaround and we might not have a need for it anyway. (I have a pending PR that makes rmw_connext_cpp allocation free, and I am currently looking at how to do the same for rmw_opensplice_cpp.)

Perhaps instead of using the ignore token, we could disable building the allocator test for fastrtps? We should probably wait on pending changes by @dirk-thomas to typesupport and how the middleware implementations are handled in CMake.

jacquelinekay:
FastRTPS is merged with passing tests on CI now, so I'm going to close this unless there are any objections. Thanks!

PR #87, opened 2015-12-15
test case for spinning before creating subscription

tfoote:
Connects to ros2/rclcpp#179

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -238,6 +238,50 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_c
   ASSERT_EQ(1, counter);
 }
 
+// Shortened version of the test for subscribing after spinning has started.
+TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), spin_before_subscription) {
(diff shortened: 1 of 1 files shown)


jacquelinekay:
This test case passes locally for Opensplice and Connext for me. Should we merge this test case?

jacquelinekay:
Eh, nevermind, doesn't pass with FastRTPS. I'll check it out.

jacquelinekay:
@tfoote, I would guess that you're encountering a problem with nondeterministic thread scheduling here.

I made a single-threaded version of this test case and the test passes on all middlewares. It makes a node, spins a bit, add a subscription, then spins some more and gets the callback.

https://github.com/ros2/system_tests/commit/bb16cadce8be8a4ad3b422e5973bee6b8cd02326

I'm guessing the `spin` thread doesn't ever execute when you run this code for some reason.

jacquelinekay:
ok, after further inspection I observe that the single-threaded case is flaky for Connext.

tfoote:
@jacquelinekay will check if this can be closed.

jacquelinekay:
we have a test case for this in master that passes now. therefore I'm going to close this.

https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_subscription.cpp#L402

PR #89, opened 2015-12-17
Fix multithreaded test by specifying publisher queue size

gerkey:
Specify a publisher queue size large enough to hold all the messages that will be published (in this case it'll be `iterations`), to avoid the possibility that in the intraprocess case we lose messages, causing the test to fail to intermittently.

I identified the problem by instrumenting (with `printf`) the intra process manager, in the place where @wjwwood had already noted that we might want to log the case where a pending message in the ring buffer gets overwritten:  https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/intra_process_manager.hpp#L254.

In the course of testing, I found these flags handy to repeat the test until it failed: `--gtest_repeat=-1 --gtest_throw_on_failure`.

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -230,19 +230,19 @@ static inline void multi_access_publisher(bool intra_process)
 
   rclcpp::executors::MultiThreadedExecutor executor;
 
-  auto pub = node->create_publisher<test_rclcpp::msg::UInt32>(node_topic_name);
+  const size_t num_messages = 5 * executor.get_number_of_threads();
+  auto pub = node->create_publisher<test_rclcpp::msg::UInt32>(node_topic_name, num_messages);
   // callback groups?
   auto msg = std::make_shared<test_rclcpp::msg::UInt32>();
   // use atomic
   std::atomic_uint timer_counter(0);
   std::atomic_uint subscription_counter(0);
 
-  const size_t iterations = 5 * executor.get_number_of_threads();
   auto timer_callback =
-    [&executor, &pub, &msg, &timer_counter, &subscription_counter, &iterations](
+    [&executor, &pub, &msg, &timer_counter, &subscription_counter, &num_messages](
     rclcpp::timer::TimerBase & timer)
     {
-      if (timer_counter.load() >= iterations) {
+      if (timer_counter.load() >= num_messages) {
         timer.cancel();
         std::atomic_uint i(0);
         // Wait for pending subscription callbacks to trigger.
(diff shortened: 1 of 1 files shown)


gerkey:
CI:

* http://ci.ros2.org/job/ci_linux/750/
* http://ci.ros2.org/job/ci_osx/638/
* http://ci.ros2.org/job/ci_windows/821/

I'm hoping that this change will fix the preexisting test failure on Windows...

gerkey:
Btw, before this change, the other `create_publisher()` function was being invoked: https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/node.hpp#L109-L114, which was getting the default queue size of 10: https://github.com/ros2/rmw/blob/master/rmw/include/rmw/qos_profiles.h#L44, which wasn't always enough, depending on how the threads got scheduled.

wjwwood:
Past me... so much smarter than current me.

I'm pretty sure @jacquelinekay and I identified this exact issue (default queue is too small) in another test. I'll see if I can find it.

wjwwood:
I think we fixed it before this test case was merged: https://github.com/ros2/system_tests/blob/7b6104df590d697503dac5b8a3b80b93d40d5c96/test_rclcpp/test/test_multithreaded.cpp#L138

But it had the nasty problem of working on my machine (4 threads * 2 < 10 (default depth)), but failed on @jacquelinekay's machine it failed (8 threads * 2 > 10 (default depth)).

gerkey:
To me, this is expected behavior from the middleware. If you plan to send a bunch of messages and need to ensure that they all get through, you need to have a big enough queue. The default is a good one, but this is a case where we need to override it.

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -230,14 +230,14 @@ static inline void multi_access_publisher(bool intra_process)
 
   rclcpp::executors::MultiThreadedExecutor executor;
 
-  auto pub = node->create_publisher<test_rclcpp::msg::UInt32>(node_topic_name);
+  const size_t iterations = 5 * executor.get_number_of_threads();
```
dirk-thomas: Should we call the variable `queue_size` here?

dirk-thomas:
Beside the small comment +1

jacquelinekay:
The Windows job looks good to me. The multithreaded test for clients fails on Connext due to ros2/rmw_connext#31, but the other cases which were previously failing pass: http://ci.ros2.org/job/ci_windows/821/testReport/(root)/test_multithreaded__/

side note:

since this problem was occurring because the 8 threads try to simultaneously push 8 things to the queue and subscriptions take up 2 queue spaces when intra-process is turned on... perhaps we could consider changing the default queue size at the rclcpp level if intra-process is turned on, or based on the `std::hardware_concurrency` of the machine? The second idea is not great because the Executor chosen might not necessarily use all cores available. But I think in the case of intra-process, the way it's implemented right now, you will always want a higher queue size than the default non intra-process size. Like the default queue size right now, these choices are somewhat arbitrary and the most important thing is to make sure this behavior is documented...

test_rclcpp/test/test_multithreaded.cpp
gerkey: Changed to `num_messages` (because it's the number of messages that we're going to send, and is used later in the code for that purpose) in b6035f5.

PR #91, opened 2015-12-17
Remove allocator test

jacquelinekay:
Connects to ros2/realtime_support#16

The allocator test is migrating to `realtime_support`.


gerkey:
+1

Issue #92, opened 2015-12-17
Intraprocess multithreaded test is flaky


gerkey:
(Not sure whether this is the same problem as #39.)

Below is the console output from a failure seen in a nightly build on OSX, running the multithreaded test with connext.

Looks like this is the sequence of events (https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_multithreaded.cpp#L78-L110):

* Publish one message, spin, and expect it to have been received by all subscribers. Instead, nobody has yet received it (failure on line 84).
* Spin more and expect that all subscribers will still have received only that first message. Instead, by this time, 1 subscriber has received it (failure on line 88).
* Reset the counter, publish 5 more messages, and expect that each subscriber will have received exactly 5 messages. Instead, both those 5 new messages and the 3 pending message from earlier are received (failure on line 110).

I'm not sure of the correct fix here. Increase the heuristic delay at the beginning (https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_multithreaded.cpp#L72)? Spin more before checking the initial delivery (https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_multithreaded.cpp#L83)? Something else?

~~~
[ RUN      ] test_multithreaded__.multi_consumer_intra_process
RTI Data Distribution Service Evaluation License issued to OSRF (OSRF01) dthomas@osrfoundation.org For non-production use only.
Expires on 08-jun-2016 See www.rti.com for more information.
/Users/osrf/jenkins/workspace/nightly_osx/ws/src/ros2/system_tests/test_rclcpp/test/test_multithreaded.cpp:84: Failure
Value of: subscriptions.size()
  Actual: 4
Expected: counter.load()
Which is: 0
callback()    1 with message data 1
/Users/osrf/jenkins/workspace/nightly_osx/ws/src/ros2/system_tests/test_rclcpp/test/test_multithreaded.cpp:88: Failure
Value of: subscriptions.size()
  Actual: 4
Expected: counter.load()
Which is: 1
callback()    1 with message data 1
callback()    2 with message data 1
callback()    3 with message data 1
callback()    4 with message data 1
callback()    5 with message data 1
callback()    6 with message data 1
callback()    7 with message data 1
callback()    8 with message data 2
callback()    9 with message data 2
callback()   10 with message data 2
callback()   11 with message data 2
callback()   12 with message data 3
callback()   13 with message data 3
callback()   14 with message data 3
callback()   15 with message data 3
[...]

PR #94, opened 2015-12-17
Fix intraprocess test failure

gerkey:
Make intraprocess more robust with a bounded sleep that checks for the goal condition after publishing (#92).

After discussing the situation with @wjwwood, we agreed that:
* there needs to be some delay between publishing and requiring that the subscriber callback has been invoked, even in the intraprocess request; and
* in lieu of any better API (e.g., waiting on a future associated with the subscriber callback), the best that we can do is to have a bounded sleep loop (which is better than a straight sleep because in the common case we'll exit early, shortening the time required for the test).

So that's what I did here in two places, plus some fixes for use of atomic variables.

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -50,7 +50,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
     [&counter, &intra_process](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
       const rmw_message_info_t & info) -> void
     {
-      ++counter;
+      counter.fetch_add(1);
       printf("callback() %4u with message data %u\n", counter.load(), msg->data);
       ASSERT_EQ(intra_process, info.from_intra_process);
     };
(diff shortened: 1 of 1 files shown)


gerkey:
Btw, I tested this fix locally on my machine under the conditions in which I was able to reproduce the failure (https://github.com/ros2/system_tests/issues/92#issuecomment-165561880), and it now works reliably.

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -50,7 +50,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
     [&counter, &intra_process](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
       const rmw_message_info_t & info) -> void
     {
-      ++counter;
+      counter.fetch_add(1);
```
jacquelinekay: as a side note, I think `fetch_add(1)` and `++counter` are actually equivalent (if they aren't, it's a bug in gcc). Using `fetch_add` makes it more obvious to the reader than the variable is atomic, however. (Same with `store` down below.)

http://en.cppreference.com/w/cpp/atomic/atomic/operator_arith
http://en.cppreference.com/w/cpp/atomic/atomic/operator%3D

gerkey: Ah, very good to know! I'll leave them as is here, but will remember in the future that the shorter syntax is just as good.

wjwwood: Cool, I didn't realize that. TIL

wjwwood:
lgtm

PR #96, opened 2015-12-18
fix missing rmw test suffix

dirk-thomas:
Before: http://ci.ros2.org/job/ci_windows/846/testReport/
After: http://ci.ros2.org/job/ci_windows/848/testReport/


gerkey:
+1

PR #98, opened 2015-12-18
Fix flaky subscription test

gerkey:
Fix flaky subscription test by adding:

* a 1ms sleep between setup and the start of publishing; and
* a maximum-2s loop of 10ms sleeps to wait for message delivery.

Both features appear to be required to ensure reliable test results with Connext when the system is under load.

These tests fail under CI most often with Connext on OSX (e.g., http://ci.ros2.org/job/ci_osx/679/testReport/%28root%29/projectroot/gtest_subscription__rmw_connext_cpp/). To reproduce locally, I load my 8-core Linux machine with `stress -c 8`. I consider the problem "fixed" when the test passes under those conditions.

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -28,6 +28,9 @@
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
+static const std::chrono::milliseconds sleep_per_loop(10);
+static const int max_loops = 200;
+
 TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning) {
   rclcpp::init(0, nullptr);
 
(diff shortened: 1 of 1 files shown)


gerkey:
FYI, the sleep-loop structure is the same as was used in #94.

wjwwood:
The changes look good, here is an OS X job with just Connext enabled (mostly to ensure that this didn't break the test, not to check that the flakiness is gone): http://ci.ros2.org/job/ci_osx/685/

In parallel I'll test it locally.

gerkey:
Sneak peak from that CI job:

~~~
22:09:59  5/31 Test  #5: gtest_subscription__rmw_connext_cpp ............................   Passed   20.70 sec
~~~

~~~
22:12:02 18/31 Test #18: gtest_subscription__rmw_connext_dynamic_cpp ....................   Passed   20.68 sec
~~~

wjwwood:
It's green. +1

PR #100, opened 2015-12-31
Fix flaky intraprocess test

gerkey:
Adjust sleeps and timeouts to be more robust, especially for Connext on OSX.

Same pattern as #98.

### DIFF
--- a/test_rclcpp/test/test_intra_process.cpp
+++ b/test_rclcpp/test/test_intra_process.cpp
@@ -21,6 +21,9 @@
 
 #include "test_rclcpp/msg/u_int32.hpp"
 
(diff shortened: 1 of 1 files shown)


gerkey:
CI:

http://ci.ros2.org/job/ci_linux/833/
http://ci.ros2.org/job/ci_osx/693/
http://ci.ros2.org/job/ci_windows/886/

I also ran the test for dozens of iterations locally, on both OSX and Linux, in both cases with a `stress -c 7` in parallel, without any failures.

esteve:
+1

gerkey:
No failures for this job from CI (that alone doesn't prove that I fixed the tests, but at least I didn't make anything worse).

PR #103, opened 2016-01-30
Spin before subscription: single-threaded

jacquelinekay:
Shows that ros2/rclcpp#179 is fixed, with one caveat--an extra sleep is necessary before publishing.

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -14,6 +14,7 @@
 
 #include <chrono>
 #include <iostream>
+#include <string>
 
 #include "gtest/gtest.h"
 
@@ -24,12 +25,42 @@
 #ifdef RMW_IMPLEMENTATION
 # define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
 # define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+
 #else
 # define CLASSNAME(NAME, SUFFIX) NAME
 #endif
 
+#define STRING_(s) #s
+#define STRING(s) STRING_(s)
+
 static const std::chrono::milliseconds sleep_per_loop(10);
-static const int max_loops = 200;
+static const size_t max_loops = 200;
+
+// Sleep for timeout ms or until a subscriber has registered for the topic
+void busy_wait_for_subscriber(
+  std::shared_ptr<const rclcpp::Node> node,
+  const std::string & topic_name,
+  std::chrono::milliseconds timeout = std::chrono::milliseconds(1),
+  std::chrono::microseconds sleep_period = std::chrono::microseconds(100))
+{
+#ifdef RMW_IMPLEMENTATION
+  if (strcmp(STRING(RMW_IMPLEMENTATION), "rmw_fastrtps_cpp") == 0) {
+    printf("FastRTPS detected, sleeping for a fixed interval\n");
+    (void)topic_name;
+    (void)node;
+    (void)sleep_period;
+    std::this_thread::sleep_for(timeout);
+    return;
+  }
+#endif
+  std::chrono::microseconds time_slept(0);
+  while (node->count_subscribers(topic_name) == 0 &&
+    time_slept < std::chrono::duration_cast<std::chrono::microseconds>(timeout))
+  {
+    std::this_thread::sleep_for(sleep_period);
+    time_slept += sleep_period;
+  }
+}
 
 TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning) {
   rclcpp::init(0, nullptr);
@@ -172,8 +203,7 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_c
     "test_subscription", callback, rmw_qos_profile_default);
 
   // wait a moment for everything to initialize
-  // TODO(gerkey): fix nondeterministic startup behavior
-  rclcpp::utilities::sleep_for(1_ms);
+  busy_wait_for_subscriber(node, "test_subscription");
 
   // start condition
   ASSERT_EQ(0, counter);
@@ -193,7 +223,7 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_c
 
   executor.spin_node_some(node);
   // spin for up to 2s
-  int loop = 0;
+  size_t loop = 0;
   while ((counter != 1) && (loop++ < max_loops)) {
     printf("callback not called, sleeping and trying again\n");
     std::this_thread::sleep_for(sleep_per_loop);
@@ -237,8 +267,7 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION),
     "test_subscription", cb_std_function, rmw_qos_profile_default);
 
   // wait a moment for everything to initialize
-  // TODO(gerkey): fix nondeterministic startup behavior
-  rclcpp::utilities::sleep_for(1_ms);
+  busy_wait_for_subscriber(node, "test_subscription");
 
   // start condition
   ASSERT_EQ(0, cb_holder.counter);
@@ -258,7 +287,7 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION),
 
   executor.spin_node_some(node);
   // spin for up to 2s
-  int loop = 0;
+  size_t loop = 0;
   while ((cb_holder.counter != 1) && (loop++ < max_loops)) {
     printf("callback not called, sleeping and trying again\n");
     std::this_thread::sleep_for(sleep_per_loop);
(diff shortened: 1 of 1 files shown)


dirk-thomas:
While this fixes the test I don't think it is feasible to expect something like that from users using the API. Therefore we have to come up with a solution which makes the API working without requiring to sleep in between.

tfoote:
Suggest removing the initial sleep and replacing it with a loop that checks for connected listeners before proceeding  into the main part of the test (where the other loop is still needed).

jacquelinekay:
https://github.com/ros2/system_tests/commit/00a134239a9102d78a65073c159ce5209bf2bb7a

I've implemented the suggestion and the Connext case passes consistently (ran it ~100 times locally). Further work includes documenting this pattern or incorporating it into the client library API via futures (in C++) or a busy loop wrapped in a function call (`await_subscription(topic name)`).

jacquelinekay:
I know the commits look a little funny right now based on a weird merge, I'm going to squash it at some point.

test_rclcpp/test/test_subscription.cpp
```diff
@@ -371,6 +371,110 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_shared_ptr_c
...
+  executor.spin_some();
+
+  auto subscriber = node->create_subscription<test_rclcpp::msg::UInt32>(
+    "spin_before_subscription", callback, rmw_qos_profile_default);
+
+  // start condition
+  ASSERT_EQ(0, counter);
+
+  size_t i = 0;
+  while (node->count_subscribers("spin_before_subscription") == 0 && ++i < max_loops) {
```
dirk-thomas: The `max_loops` logic is wrong. E.g. if `max_loops = 1` the while loop won't run at all.

dirk-thomas: Also while not being defined by this PR `max_loops` is a constant and should therefore be upper case.

wjwwood: Actually constants should be all lower case and underscores with a `k` prefix: https://google.github.io/styleguide/cppguide.html#Constant_Names

We don't have an exception for this in out style guide either: https://github.com/ros2/ros2/wiki/Developer-Guide#c-1

Macros and preprocessor defines are the only thing that should be all uppercase I think.

wjwwood: Sorry, they would be `kCamelCase` with a `k` suffix according to the GSG, but I'm not sure if that is covered by our exception where we use all lower and underscore for functions and variables.

jacquelinekay: I've redone this logic a little bit by putting it into a function and using a more readable counting method for the busy wait. I'm also reusing the function throughout the tests in this file to wait for a subscription to register for a certain topic, rather than waiting a fixed amount of time.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/908
http://ci.ros2.org/job/ci_osx/760
http://ci.ros2.org/job/ci_windows/986

test_rclcpp/test/test_subscription.cpp
jacquelinekay: I don't know about the style for `max_loops`, but what bothered me more is that it's signed type and really should be an unsigned type. So I also changed it to size_t and all of the counting variables throughout these tests to size_t.

jacquelinekay:
darn, I realized this is going to fail on FastRTPS because count_subscribers is not implemented for it yet.
```
C++ exception with description "could not count subscribers: not implemented, at /home/jackie/code/ros2_ws/src/eProsima/ROS-RMW-Fast-RTPS-cpp/rmw_fastrtps_cpp/src/functions.cpp:1803" thrown in the test body.
```

jacquelinekay:
So, from briefly looking at rmw_fastrtps_cpp, a nontrivial amount of new code is needed to implement `count_subscribers` and `count_publishers`. I can look into either completing this feature or changing the functions so that they don't throw exceptions and return 0, so that FastRTPS simply times out in these tests.

gerkey:
From our discussion: ticket upstream to implement the missing call and for now call it, catching the exception, and then sleeping some heuristic amount of time.

jacquelinekay:
Ticketed in eProsima/ROS-RMW-Fast-RTPS-cpp#14

jacquelinekay:
Added some preprocessor logic to ignore FastRTPS in 8e966dd.

CI for good measure:

http://ci.ros2.org/job/ci_linux/917
http://ci.ros2.org/job/ci_osx/778
http://ci.ros2.org/job/ci_windows/1017

wjwwood:
https://github.com/ros2/system_tests/commit/8e966dddfa0d7483416455d79a45a7c0e74c95a0 lgtm

jacquelinekay:
Style fixes in 9557d67

jacquelinekay:
The linux build failed because of a problem with another package, so I'm rerunning it here (also to show the fixes for the linter).

http://ci.ros2.org/job/ci_linux/921/

jacquelinekay:
@wjwwood Is your "lgtm" for that commit or for the whole PR?

dirk-thomas:
This still needs a passing linux CI job.

jacquelinekay:
The passing Linux CI job is here:

http://ci.ros2.org/job/ci_linux/925/

test_rclcpp/test/test_subscription.cpp
```diff
@@ -24,12 +25,44 @@
...
+// Sleep for timeout ms or until a subscriber has registered for the topic
+void busy_wait_for_subscriber(
+  std::shared_ptr<const rclcpp::Node> node,
+  const std::string & topic_name,
+  std::chrono::milliseconds timeout = std::chrono::milliseconds(1),
+  std::chrono::microseconds sleep_period = std::chrono::microseconds(100))
+{
+  std::chrono::microseconds time_slept(0);
+#ifdef RMW_IMPLEMENTATION
+  if (strcmp(xstr(RMW_IMPLEMENTATION), "rmw_fastrtps_cpp") == 0) {
```
dirk-thomas: Can `xstr(RMW_IMPLEMENTATION)` be simplified to `#RMW_IMPLEMENTATION`?

test_rclcpp/test/test_subscription.cpp
```diff
@@ -24,12 +25,44 @@
...
+#else
+  if (true) {
```
dirk-thomas: Why `if (true)`? Just an open brace `{` should be ok.

test_rclcpp/test/test_subscription.cpp
```diff
@@ -24,12 +25,44 @@
...
+#endif
+    printf("FastRTPS detected, sleeping for a fixed interval");
```
dirk-thomas: MIssing newline.

test_rclcpp/test/test_subscription.cpp
dirk-thomas: Never mind. With the later `else` it needs an `if`. But should it be `false` if `RMW_IMPLEMENTATION` is unknown?

jacquelinekay: What do you think of how I restructured things in 7a1f2db? I added a new macro to save code, though I admit it looks a little strange.

test_rclcpp/test/test_subscription.cpp
jacquelinekay: no, unfortunately, two levels of macros is necessary. `#RMW_IMPLEMENTATION` does not compile, and with only one level macro the test does not pass.

https://gcc.gnu.org/onlinedocs/cpp/Stringification.html

dirk-thomas:
The recent job http://ci.ros2.org/job/ci_linux/925/ seems to not test this branch if I interpret the output of the vcs commands correctly??

dirk-thomas:
I triggered http://ci.ros2.org/job/ci_linux/928/ to actually test this branch. I am not sure what you put in as a branch name but it contains a unicode character in the middle of it.

jacquelinekay:
thanks. I seem to have a problem where when I type branch names or copy/paste branch names from my terminal, random invisible unicode characters are inserted into the branch name and are thus unable to be parsed by Jenkins. I should probably figure out why that happens and when exactly it happens. It's annoying because it never renders in my browser. Maybe I'm just bad at typing, too bad it's what I do for a living.

test_rclcpp/test/test_subscription.cpp
```diff
@@ -24,12 +25,42 @@
...
+static const size_t max_loops = 200;
+
+// Sleep for timeout ms or until a subscriber has registered for the topic
+void busy_wait_for_subscriber(
+  std::shared_ptr<const rclcpp::Node> node,
+  const std::string & topic_name,
+  std::chrono::milliseconds timeout = std::chrono::milliseconds(1),
+  std::chrono::microseconds sleep_period = std::chrono::microseconds(100))
+{
+  std::chrono::microseconds time_slept(0);
```
dirk-thomas: This line could be moved below the `ifdef` block.

dirk-thomas:
Forget about the previous CI job results. They all failed to find Connext.

dirk-thomas:
New job: http://ci.ros2.org/job/ci_linux/936/

dirk-thomas:
+1 to squash and merge.

PR #106, opened 2016-02-11
Fix race condition in test_executor

gerkey:
The `multiple_executors` test case was sometimes timing out: http://ci.ros2.org/view/nightly/job/nightly_osx/213/testReport/.

In manual testing, I observed that the counter shared between the two threads was sometimes skipping over the goal value of 20, so that neither executor thread would return (they each require the counter to equal the goal value). This is presumably because sometimes the counter increment happens back-to-back in [callback1](https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_executor.cpp#L83) and [callback2](https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_executor.cpp#L97) without giving either one the chance to compare the counter to the goal value.

This PR gives each thread its own counter, which avoids that race condition. In my local testing on OSX, the test now passes reliably when run repeatedly, including when the system is under heavy load.

Given the simplicity of the change, I'm inclined to skip CI jobs. Let me know if you disagree.

### DIFF
--- a/test_rclcpp/test/test_executor.cpp
+++ b/test_rclcpp/test/test_executor.cpp
@@ -69,18 +69,20 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), multithreaded_spin_call) {
 
 // Try spinning 2 single-threaded executors in two separate threads.
 TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), multiple_executors) {
-  std::atomic_uint counter;
-  counter = 0;
+  std::atomic_uint counter1;
+  counter1 = 0;
+  std::atomic_uint counter2;
+  counter2 = 0;
   const uint32_t counter_goal = 20;
 
   // Initialize executor 1.
   rclcpp::executors::SingleThreadedExecutor executor1;
-  auto callback1 = [&counter, &counter_goal, &executor1]() {
-      if (counter == counter_goal) {
+  auto callback1 = [&counter1, &counter_goal, &executor1]() {
+      if (counter1 == counter_goal) {
         executor1.cancel();
         return;
       }
-      ++counter;
+      ++counter1;
     };
   auto node1 = rclcpp::Node::make_shared("multiple_executors_1");
   auto timer1 = node1->create_wall_timer(1_ms, callback1);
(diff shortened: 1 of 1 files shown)


jacquelinekay:
Thank you for debugging this!

I've launched CI on OSX that will repeat only the Executor test 20 times (if I remembered my ctest syntax correctly).

http://ci.ros2.org/job/ci_osx/795/

gerkey:
@jacquelinekay That's some advanced ctest usage. Thanks!

dirk-thomas:
LGTM.

gerkey:
As far I can tell, @jacquelinekay's CI job only ran the `rmw_connext_cpp` flavor of that test (maybe the `--test-regex` flag needs different input?). But that test did run repeatedly and pass.

jacquelinekay:
hmm. That's not good--the "Use Opensplice" and "Use FastRTPS" flags were on in the Jenkins parameters. Are we suffering from more CI issues?

gerkey:
As far I can see from the console output all flavors of that test were built, but only the `rmw_connext_cpp` flavor was run.

dirk-thomas:
This build should test all the rmw impl.: http://ci.ros2.org/job/ci_osx/796/parameters/

dirk-thomas:
The test looks good for all rmw impl.: http://ci.ros2.org/job/ci_osx/796/testReport/(root)/

jacquelinekay:
Interesting. Perhaps Brian had not rebased against a recent master with the generator expression change. Regardless, I'm not too worried about it anymore.

dirk-thomas:
The console output captures what happens: http://ci.ros2.org/job/ci_osx/795/consoleFull#console-section-5 Also the CI jobs rebase any custom branch against the default branch to ensure to only test the merged state. So this has nothing to do with an outdated branch.

I can reproduce the same weird regex matching behavior locally. But when I try the same for a minimal example the regex matches differently (how it should in the first place). I think this needs investigation why "our" CTests behave differently when being selected using a regex.

gerkey:
Great to know that the tests pass on all implementations. Thanks for the help with testing.

dirk-thomas:
The problem with the command `ctest -R gtest_executor*` is that it is being executed in the build folder of the package. The shell will then expand it into `ctest -R gtest_executor__rmw_connext_cpp gtest_executor__rmw_connext_dynamic_cpp gtest_executor__rmw_fastrtps_cpp gtest_executor__rmw_opensplice_cpp`. Since `-R` only cares about the first argument only the first test gets executed.

Consequently all of the following variations work as expected:
* `ctest -R ^gtest_executor*`
* `ctest -R "gtest_executor*"`
* `ctest -R gtest_executor\*`

The only improvement I could imagine would be to invoke `ctest` in an empty folder to ensure that the shell can't expand the arguments intended as a regex. But afaik `ctest` does not support being invoked in a different folder then the build folder.

wjwwood:
`zsh` has a command called `noglob` which you can put before your entire line and it will prevent the shell from considering characters like `*`.

dirk-thomas:
`set -o noglob` seems to be supported even in plain shell. Can you confirm that it works with zsh, too?

wjwwood:
Yeah, it does.

dirk-thomas:
I considered adding this to `ament_tools` but I think it would be even more confusing if the same arguments behave differently when ran with `ament test` and `ctest`. So I think we should keep it as is and only be aware of it.

PR #107, opened 2016-02-18
Add test for listing all parameters

gerkey:
Connects to ros2/rclcpp#196.

Test that we can list all parameters by passing an empty prefix list.


gerkey:
Replaced by the same change in #140.

PR #108, opened 2016-03-04
changed how the subscription_and_spinning test works

wjwwood:
So I've attempted to make this test less flaky. What I did was use a `std::shared_future` and `std::promise` so that I could have the executor spin until the predicate was fulfilled. The predicate was that the subscription's callback was called. So I replaced the use of `spin_once` with `spin_until_future_complete`.

### DIFF
--- a/test_rclcpp/test/test_subscription.cpp
+++ b/test_rclcpp/test/test_subscription.cpp
@@ -62,6 +62,25 @@ void busy_wait_for_subscriber(
   }
 }
 
+template<typename DurationT>
+void wait_for_future(
+  rclcpp::executor::Executor & executor,
+  std::shared_future<void> & future,
+  DurationT & timeout)
+{
+  using rclcpp::executor::FutureReturnCode;
+  rclcpp::executor::FutureReturnCode future_ret;
+  auto start_time = std::chrono::steady_clock::now();
+  future_ret = executor.spin_until_future_complete(future, timeout);
+  auto elapsed_time = std::chrono::steady_clock::now() - start_time;
+  // *INDENT-OFF*
+  EXPECT_EQ(FutureReturnCode::SUCCESS, future_ret)
+    << "future failed to be set after: "
+    << std::chrono::duration_cast<std::chrono::milliseconds>(elapsed_time).count()
+    << " milliseconds\n";
+  // *INDENT-ON*
+}
+
 TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning) {
   rclcpp::init(0, nullptr);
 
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/test_subscription.cpp
```diff
@@ -129,53 +156,46 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
...
+    sub_called = std::promise<void>();
+    sub_called_future = sub_called.get_future();
+    wait_for_future(executor, sub_called_future, 10_ms);
+    ASSERT_EQ(4, counter);
+
+    // check for last pending call (blocking)
+    printf("spin_until_future_complete() - callback (5) expected\n");
+    sub_called = std::promise<void>();
+    sub_called_future = sub_called.get_future();
+    wait_for_future(executor, sub_called_future, fail_after_timeout);
```
wjwwood: I replaced the use of `spin_some` here because I believe it is not something we can reliably check. Basically we cannot know that the time since publishing "4" and "5" has been long enough for both to be ready to be taken in one `spin_some` call.

test_rclcpp/test/test_subscription.cpp
```diff
@@ -129,53 +156,46 @@ TEST(CLASSNAME(test_subscription, RMW_IMPLEMENTATION), subscription_and_spinning
...
     ASSERT_EQ(5, counter);
   }
   // the subscriber goes out of scope and should be not receive any callbacks anymore
 
-  // wait a moment for everything to initialize
-  // TODO(gerkey): fix nondeterministic startup behavior
-  rclcpp::utilities::sleep_for(1_ms);
-
```
wjwwood: I removed this because even if the underlying middleware is still connected and the `rmw` subscription gets notified as ready for the "6" we publish below, it should never make it to the `rclcpp` subscription callback since that class has been destroyed.

wjwwood:
If we're satisfied with the approach in this pr then I can give a similar treatment to the other tests.

tfoote:
waiting for the event makes more sense than expecting it to always finish in one cycle. 

We should also assert the wait does not timeout in most places.

wjwwood:
@tfoote checking for the timeout is what it should be doing: https://github.com/ros2/system_tests/pull/108/files#diff-783ac165669f5ffa237b192404f75f4eR76

If you see some place we're not checking that let me know.

tfoote:
Sorry, I didn't look inside that function for the checks. I'd expected it to be a more generic function that might be promotable as a reusable utility function in user code vs just being a test function. And in the test we would check the return code instead of testing inside the function.

dirk-thomas:
Except that `wait_for_future` should probably go into `rclcpp` this lgtm.

wjwwood:
@tfoote I don't understand what you're suggesting exactly and you give a code snippet example or describe it more? One note is that functions that use gtest macros must return void. 

@dirk-thomas where in rclcpp should that function go? It asserts conditions using gtest.

dirk-thomas:
The `EXPECT_EQ` itself should not be part of the generic function. That needs to be checked in the test instead. But the rest of the function is generic.

wjwwood:
But the purpose of the function is to deduolicate the code that does the checking, otherwise it is just calling executor.spin_until_future_complete. I do not believe this is a useful function for users.

wjwwood:
Bump, does anyone still think the `wait_for_future` should be refactored out of the tests?

@tfoote do you have any more details on your suggestion from before?

tfoote:
@wjwwood I was thinking along the same lines as @dirk-thomas but if it's not that useful outside the test it's fine as is.

dirk-thomas:
After discussing that this will only be used for these tests and other tests will simply use two lines of the function directly, lgtm.

esteve:
I'd keep the `wait_for_future` function in the tests, and if it proves useful in `rclcpp` I'd move it there when so.

wjwwood:
Ok, running CI (linux to start) for this one since the changes are non-trivial (including `rclcpp` changes):

- http://ci.ros2.org/job/ci_linux/1074/

wjwwood:
I call shenanigans on this warning:

- http://ci.ros2.org/job/ci_linux/1074/warnings17Result/new/
- SO: http://stackoverflow.com/a/18680397/671658

Also it doesn't produce a warning in clang.

Also if I comment out one of the cases I get this warning:

```
/Users/william/ros2_ws/src/ros2/rclcpp/rclcpp/src/rclcpp/executor.cpp:589:11: warning: enumeration value 'TIMEOUT' not handled in switch [-Wswitch]
  switch (future_return_code) {
          ^
1 warning generated.
```

I fixed it (hopefully) anyways :pouting_cat:: ros2/rclcpp@a8f2b82

wjwwood:
There are also a tonne of test failures, so I'm going to look into those.

wjwwood:
I fixed a major bug where the default case, `timeout == -1`, did not work correctly in ros2/rclcpp@d73b86a. Here's another Linux CI:

- http://ci.ros2.org/job/ci_linux/1075/

wjwwood:
I cannot reproduce the failure on the farm locally. So, I moving the tests in `rclcpp` to `test_rclcpp` since there is already a `test_spin` test that creates a version for each middleware (which I think could be related).  Running a new CI to see if that gives me more info:

- http://ci.ros2.org/job/ci_linux/1077/

wjwwood:
It turns out that `std::async` has different default behavior on Linux and OS X, so in bb5cebd I enforced it to be `std::launch::async`.

wjwwood:
New Linux CI:

- http://ci.ros2.org/job/ci_linux/1079/

wjwwood:
That last linux build had a pair of failures related to the pendulum tests. @jacquelinekay can you have a look at them and see if it is flaky or it might have something to do with my changes?

wjwwood:
Here is a full suite of CI:

- http://ci.ros2.org/job/ci_linux/1080/
- http://ci.ros2.org/job/ci_osx/876/
- http://ci.ros2.org/job/ci_windows/1137/

jacquelinekay:
that's pretty odd. Your changes in rclcpp should affect that executable since it doesn't use `spin_until_future_complete`. I will investigate.

wjwwood:
should => shouldn't?

jacquelinekay:
right

wjwwood:
Thanks for looking at it, we can see if it persists in http://ci.ros2.org/job/ci_linux/1080/ (about 4 minutes out)

wjwwood:
@jacquelinekay http://ci.ros2.org/job/ci_linux/1080/ is clean and didn't have any changes from http://ci.ros2.org/job/ci_linux/1079/, fyi.

wjwwood:
Currently I'm still working out why these two tests:

- http://ci.ros2.org/job/ci_windows/1138/testReport/projectrootC_.J.workspace.ci_windows.ws.src.ros2.system_tests/test_rclcpp/gtest_spin__rmw_opensplice_cpp/
- http://ci.ros2.org/job/ci_windows/1138/testReport/projectrootC_.J.workspace.ci_windows.ws.src.ros2.system_tests/test_rclcpp/gtest_executor__rmw_opensplice_cpp/

Sometimes block until timeout on Windows. In this job, for example, those same tests did not fail:

http://ci.ros2.org/job/ci_windows/1144/testReport/

But locally on my VM they always seem to stall out.

wjwwood:
Ok, I narrowed down the remaining flaky tests to be a duplicate of something @tfoote is working on. I couldn't find an issue just yet. @tfoote is there one I can link to?

Either way, this has +1's on both PR's and a clean CI (excluding the known flaky and failing tests on Windows), so I'll squash and merge it.

jacquelinekay:
sounds good to me

PR #109, opened 2016-03-04
fix comparison warnings within uses of gtest macros

wjwwood:
There were a lot of warnings when building the `test_rclcpp` package with `RelWithDebInfo` on OS X. Most are comparison between signed/unsigned numbers via (within) gtest macros.

PR #110, opened 2016-03-14
resolve cpplint warnings

wjwwood:
See: https://github.com/ros2/system_tests/pull/109#issuecomment-195020801


jacquelinekay:
+1

wjwwood:
Here is a OpenSplice only Linux CI job which shows them as fixed:

- http://ci.ros2.org/job/ci_linux/1093/

PR #111, opened 2016-03-14
fix the order of the assert_ge check in test_publisher

wjwwood:
Looks like I got the order wrong on an `ASSERT_GE` when addressing the cpplint warnings. See: https://github.com/ros2/system_tests/pull/109#issuecomment-195020801


tfoote:
+1

wjwwood:
I also fixed a printf warning that only happens on Linux (chrono returns `long long int` on my machine and `long int` on the Linux CI job...). I just pushed it.

Here is CI that hopefully shows it is fixed on Linux (OpenSplice only):

- http://ci.ros2.org/job/ci_linux/1095/

wjwwood:
I had to fix another style warning.

wjwwood:
This one is clean:

- http://ci.ros2.org/job/ci_linux/1097/

PR #112, opened 2016-03-17
Add tests for notify guard condition in node

jacquelinekay:
Connects to ros2/rclcpp#209

### DIFF
--- a/test_rclcpp/test/test_executor.cpp
+++ b/test_rclcpp/test/test_executor.cpp
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 #include <atomic>
+#include <cinttypes>
+#include <future>
 #include <stdexcept>
 
 #include "gtest/gtest.h"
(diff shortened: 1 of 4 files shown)


jacquelinekay:
I find this test is currently flaky for Connext but no other platform. Looking into it.

test_rclcpp/test/test_executor.cpp
```diff
@@ -118,6 +121,84 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), multiple_executors) {
...
+    auto timer = node->create_wall_timer(
+      1_ms,
+      [&executor, &timer_triggered](rclcpp::TimerBase & timer)
+    {
+      timer_triggered = true;
+      executor.cancel();
+      timer.cancel();
+    });
+
+    spin_thread.join();
```
wjwwood: I'm not sure it's worth changing this test, but for the future you might consider using a future here rather than the `timer_triggered` boolean. Something like this:

```c++
std::thread spin_thread(executor_spin_lambda);
std::promise<void> timer_triggered_promise;
std::shared_future<void> timer_triggered_future(timer_triggered_promise.get_future());
auto timer = node->create_wall_timer(
  1_ms,
  [&executor, &timer_triggered_promise](rclcpp::TimerBase & timer)
{
  timer_triggered_promise.set_value();
  timer.cancel();
});

EXPECT_EQ(std::future_status::ready, timer_triggered_future.wait_for(100_ms));
executor.cancel();
spin_thread.join();
```

The advantage of this approach is that if the timer fails to trigger the test doesn't just hang, but instead it fails an expected check and the test can continue to try other things. Currently the only way this test fails is if the test timeout kills the test.

test_rclcpp/test/test_executor.cpp
```diff
@@ -118,6 +121,84 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), multiple_executors) {
...
+    auto future_result = client->async_send_request(request);
+    EXPECT_EQ(future_result.get()->sum, 6);
+    executor.cancel();
+    spin_thread.join();
+  }
+}
+
+// test removing a node
+
+// test notify with multiple nodes
```
wjwwood: Do you intend to fill these tests in, or should these become TODO's?

jacquelinekay: These are notes about corner cases that I thought of and didn't write because I was lazy. Better tests are usually a good thing, but waiting on me to write them will slow down progress. Do you think I should take another stab at them?

wjwwood: I have similarly merged things before I had what I considered complete coverage. I would say it's up to you. You can just put in TODO's / make an issue to track them (so we don't forget about them) or you can try to implement them now (maybe you have a clearer picture of how to do so right now rather than in the future).

jacquelinekay:
The test was flaky on Connext; I added a busy_wait_for_subscriber and decided to put it in its own file to share it between multiple tests: 35720eabea2db589d3b3dc9a96f5b3c7c64ac22d

PR #113, opened 2016-04-05
Use rcl

jacquelinekay:
Connects to ros2/rclcpp#204

`rclcpp::init` is actually required now. It also can't be called twice.


wjwwood:
These changes look good to me too.

PR #116, opened 2016-05-03
add classname label to some tests

jacquelinekay:


wjwwood:
Thanks @jacquelinekay lgtm.

PR #117, opened 2016-05-04
add message which has fields with the same non-primitive type

dirk-thomas:
Add a message which has multiple fields with the same non-primitive type.

* http://ci.ros2.org/job/ci_linux/1263/
* http://ci.ros2.org/job/ci_osx/1009/
* http://ci.ros2.org/job/ci_windows/1296/

This is mainly to demonstrate that ros2/rmw_opensplice#109 is not necessary. Should we still merge this (to catch potential problems in the future) or just close it?

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -28,6 +28,7 @@ if(BUILD_TESTING)
     "msg/DynamicArrayNested.msg"
     "msg/DynamicArrayPrimitives.msg"
     "msg/Empty.msg"
+    "msg/FieldsWithSameType.msg"
     "msg/Nested.msg"
     "msg/Primitives.msg"
     "msg/StaticArrayNested.msg"
--- a/test_communication/msg/FieldsWithSameType.msg
+++ b/test_communication/msg/FieldsWithSameType.msg
@@ -1,0 +1,2 @@
+Primitives primitive_values1
+Primitives primitive_values2


mikaelarguedas:
I think that it can't hurt to merge it and can be useful in the future

dirk-thomas:
Needed to be registered separately since it doesn't have a pub / sub test: 2787130b6edf68736ab34bb0b1259761c8140fa6

wjwwood:
Should we close https://github.com/ros2/rmw_opensplice/pull/109?

mikaelarguedas:
@dirk-thomas If the goal of this new message type is only to test message generation and not use it for communication testing, should we move that in another package lower in the stack? looks like it could go in `rosidl_generator_py` given that it relies on c type support. Or do you see a better candidate to host this message?

dirk-thomas:
While it doesn't have any fixtures it would be good to ensure that every language can handle this case.

mikaelarguedas:
So (just to confirm my understanding) it should be in a package downstream of `rosidl_generator_default`, because we want to be sure that it can be generated/compiled by all generators? or should we have a test trying to use this message in every supported language?

dirk-thomas:
Both has its pros and cons. If each generator would ensure this that would be great. If we do it in a downstream package we at least ensure that all generators pass - but this only gets tested when that package get build. In each generator it would immediately get checked when the generator is modified and the generator can do some additional language specific checks.

mikaelarguedas:
Given that we start adding msg definitions for more corner cases ([including fields with same type](https://github.com/ros2/rosidl/pull/236/files#diff-8cdf70893439c4b2d68ed248e6aa0369R1)) in the generators themselves; we should be able to remove it from here once it's propagated to all message generators. @clalancette FYI

PR #118, opened 2016-05-09
extend test to cover another case

dirk-thomas:
Unit test for ros2/rclcpp#219.

Currently fails when the publisher and subscriber executables are being invoked separately. The jobs doesn't fail (http://ci.ros2.org/job/ci_linux/1268/) since the multi executable tests are not run: see #119


tfoote:
+1

PR #119, opened 2016-05-10
fix generation of tests with multiple executables

dirk-thomas:
Fix wrong CMake variable to reenabled tests with multiple executables. Numerous of the tests are currently failing:

* http://ci.ros2.org/job/ci_linux/1269/
* http://ci.ros2.org/job/ci_osx/1015/
* http://ci.ros2.org/job/ci_windows/1303/

PR #120, opened 2016-05-17
Ensure using correct rmw implementation in tests

dhood:
Connects to ros2/examples#96

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -76,7 +76,7 @@ if(BUILD_TESTING)
       COMMAND "${target_path}/${target}${target_suffix}" empty
       TIMEOUT 15
       GENERATE_RESULT_FOR_RETURN_CODE_ZERO
-    )
+      ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
     set_tests_properties(
       ${target}${target_suffix}
       PROPERTIES REQUIRED_FILES "${target_path}/${target}${target_suffix}"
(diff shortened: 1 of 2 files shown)


test_communication/CMakeLists.txt
```diff
@@ -123,7 +123,8 @@ if(BUILD_TESTING)
         INPUT "${CMAKE_CURRENT_BINARY_DIR}/test_publisher_subscriber_cpp${test_suffix}.py.configured"
       )
 
-      ament_add_nose_test(publisher_subscriber_cpp${test_suffix} "${CMAKE_CURRENT_BINARY_DIR}/test_publisher_subscriber_cpp${test_suffix}_$<CONFIGURATION>.py" TIMEOUT 15)
+      ament_add_nose_test(publisher_subscriber_cpp${test_suffix} "${CMAKE_CURRENT_BINARY_DIR}/test_publisher_subscriber_cpp${test_suffix}_$<CONFIGURATION>.py" TIMEOUT 15
+        ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
```
dirk-thomas: This test invokes two binaries which might use different RMW impl. therefore a single environment variable can't be used to test.

This needs to be reverted. Same below for the `requester_replier_*` tests.

dirk-thomas: I will revert these two changes in 412319ea683385ce259f73f70e7c86636447c2d8 which is part of #119.

wjwwood: I guess the `rmw_implementation` needs to be passed down into the `.py` file and then the env be set by `launch`.

Can we just roll that change into the pr that re-enables all the tests that were previously disabled?

mikaelarguedas: agreed, all the reenabled tests will fail without that change, so roll that change in https://github.com/ros2/system_tests/pull/119 seems to be the way to go.

dirk-thomas: For now I will remove the check in #119. The check can be added afterwards in a separate PR.

dhood: @wjwwood did I hear you say you're working on this? otherwise I'll fix it

wjwwood: I got it, thanks.

dhood: ok thank you!

wjwwood: https://github.com/ros2/system_tests/pull/126

PR #122, opened 2016-05-28
add a bunch of tests for rcl and rosidl_generator_c messages

jacquelinekay:
using the test_communication message definitions.

Not really ready for review yet.

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -109,13 +109,6 @@ if(BUILD_TESTING)
   endfunction()
 
   function(custom_test_c target)
-    # TODO(jacquelinekay): remove these when C type support is supported by typesupport_introspection.
-    #                after removing, we can also recombine the function and macro
-    if("${rmw_implementation} " STREQUAL "rmw_fastrtps_cpp ")
-      message(STATUS "Skipping tests for '${rmw_implementation}'")
-      return()
-    endif()
-
     # get typesupport of rmw implementation to include / link against the corresponding interfaces
     get_rmw_typesupport(typesupport_impls "${rmw_implementation}" LANGUAGE "c")
 
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -128,7 +128,6 @@ if(BUILD_TESTING)
       ${PROJECT_NAME} rosidl_generator_c)
     ament_target_dependencies(${target}${target_suffix}
       "rcl${target_suffix}")
-    # register executable as a test
     get_target_property(target_path ${target}${target_suffix}
       RUNTIME_OUTPUT_DIRECTORY)
     if(NOT target_path)
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -108,6 +108,65 @@ if(BUILD_TESTING)
       "rclcpp${target_suffix}")
   endfunction()
 
+  function(custom_test_c target)
+    # TODO(jacquelinekay): remove these when C type support is supported by typesupport_introspection.
+    #                after removing, we can also recombine the function and macro
+    if("${rmw_implementation} " STREQUAL "rmw_fastrtps_cpp ")
+      message(STATUS "Skipping tests for '${rmw_implementation}'")
+      return()
+    endif()
+
+    # get typesupport of rmw implementation to include / link against the corresponding interfaces
+    get_rmw_typesupport(typesupport_impls "${rmw_implementation}" LANGUAGE "c")
+
+    # add_executable(${target}${target_suffix} ${ARGN})
+    ament_add_gtest(
+      "${target}${target_suffix}" ${ARGN}
+      TIMEOUT 15
+      ENV RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation})
+    target_link_libraries(${target}${target_suffix}
+      ${_AMENT_EXPORT_ABSOLUTE_LIBRARIES}
+      ${_AMENT_EXPORT_LIBRARY_TARGETS})
+    add_dependencies(${target}${target_suffix} ${PROJECT_NAME})
+    foreach(typesupport_impl ${typesupport_impls})
+      rosidl_target_interfaces(${target}${target_suffix}
+        ${PROJECT_NAME} ${typesupport_impl})
+    endforeach()
+    rosidl_target_interfaces(${target}${target_suffix}
+      ${PROJECT_NAME} rosidl_generator_c)
+    ament_target_dependencies(${target}${target_suffix}
+      "rcl${target_suffix}")
+    # register executable as a test
+    get_target_property(target_path ${target}${target_suffix}
+      RUNTIME_OUTPUT_DIRECTORY)
+    if(NOT target_path)
+      set(target_path "${CMAKE_CURRENT_BINARY_DIR}")
+    endif()
+    set_tests_properties(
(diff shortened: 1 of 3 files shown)


mikaelarguedas:
It might be worth it to mutualise effort on this. I started refactoring
this to match the other test fixtures (CPP and Python) and debug the array
failures with fastrtps and type support introspection c

dirk-thomas:
Do these tests cover the case from eProsima/ROS-RMW-Fast-RTPS-cpp#45? If yes it would be good to run two CI builds with it - one before the patch and one after.

mikaelarguedas:
For now it doesn't but it will cover all cases after the refactor.
The python tests do cover all cases, all the array messages tests are failing for fastrtps.
I tested https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/45 locally, it fixes all the dynamic primitive arrays (`BoundedArrayPrimitives` and `DynamicArrayPrimitives`) but not the other cases (`BoundedArrayNested`,`DynamicArrayNested`, `StaticArrayNested`, `StaticArrayPrimitives`).

Detail of test failures:
BoundedArrayNested: segfault
StaticArrayNested: segfault
DynamicArrayNested: message received is empty
StaticArrayPrimitives: all fields are populated correctly except the string arrays that contain only empty strings

mikaelarguedas:
@ros2/team I would need some input on this.

**Context:**
 This PR tests all `test_communication` messages with C typesupport at the rcl level.
Tests are passing but it needs some love to be sure the tests won't end up being flaky on slow machine (a waitset basically).

**Concern:**
 Not sure what the purpose of this PR is today, all messages with C typesupport are already tested by the Python tests in this package.

**Pros:**
 * If we extend the ROS IDL feature set it could come in handy to test them at a lower level than rclpy

**Cons:**
 * The Python tests will have to be extended to test any new feature anyway, making this redundant, they would expose the same failures and won't be much harder to introspect
 * Adding these tests will take more time on the buildfarm without providing any additional information about the state of the system

**Question:**
Should I spend more time on this or should we put it in backlog ?

wjwwood:
My personal opinion is that the Python tests would ideally cover this case for us, making these tests redundant, at least for now. At the time this was originally opened, I think rclpy was not in a place where it could test these C messages effectively. I think that is perhaps not the case any longer. I'd vote to spend time elsewhere unless there is a compelling need to have these tests in addition to the Python tests.

dirk-thomas:
Since this PR implements a lot of tests on the C level (which is preferable over testing the same stuff from Python which sits another level above) I would consider merging this. The main question for me is what is the status of this PR? Are the any of the tests currently flaky? If yes, can they be updated to simply wait "a bit"? While that might not be the final great solution it might be enough for now to merge this...

mikaelarguedas:
They haven't proven to be flaky yet, they passed on [linux](http://ci.ros2.org/job/ci_linux) and [osx](http://ci.ros2.org/job/ci_osx/1439) on the first try (windows CI failed this morning for unrelated reason).
I ran them in repeated mode on my laptop 20 times without seeing a single failure.
The potential flakyness is due to the design of the tests themselves (waiting one sec and assume publisher and subscriber are up and connected). this 1 sec period makes these tests pretty slow (~1min for connext, ~20sec for fastrtps).
I can iterate on making them faster and more reliable if we reach consensus on their usefullness

mikaelarguedas:
or merge them as is and iterate on them when/if they become flaky

dirk-thomas:
I would say the later is good enough. Making them a bit faster would be nice if it doesn't take too much effort but other stuff might be more important atm.

mikaelarguedas:
* Rebased
 * Added waitset for both connection established and message reception

Tests now takes ~30sec for connext/connext_dynamic, <3 sec for opensplice and fastrtps

Tests failures fixed in #172 

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1890)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1441/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1808/)

mikaelarguedas:
rebased and proper graph guard condition:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1891)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1442/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1810/)

test_communication/CMakeLists.txt
```diff
@@ -108,6 +108,38 @@ if(BUILD_TESTING)
...
+    add_dependencies(${target}${target_suffix} ${PROJECT_NAME})
+    foreach(typesupport_impl ${typesupport_impls})
+      rosidl_target_interfaces(${target}${target_suffix}
+        ${PROJECT_NAME} ${typesupport_impl})
+    endforeach()
+    rosidl_target_interfaces(${target}${target_suffix}
+      ${PROJECT_NAME} rosidl_generator_c)
+    ament_target_dependencies(${target}${target_suffix}
+      "rcl${target_suffix}")
+    # register executable as a test
```
dirk-thomas: The following lines are not doing what the comment describes?

test_communication/test/test_messages_c.cpp
```diff
@@ -0,0 +1,1127 @@
...
+      msg->float64_value = 1;
+      msg->int8_value = 1;
+      msg->uint8_value = 1;
+      msg->int16_value = 1;
+      msg->uint16_value = 1;
+      msg->int32_value = 1;
+      msg->uint32_value = 1;
+      msg->int64_value = 1;
+      msg->uint64_value = 1;
+      char string_value[200] = {};
```
dirk-thomas: The tests in other languages explicitly try a larger size since <= 255 might work but not for higher numbers.

test_communication/CMakeLists.txt
mikaelarguedas: artefact indeed, removed in https://github.com/ros2/system_tests/pull/122/commits/09b6a18f765d78ac8e2035c982df9135a109b725

test_communication/test/test_messages_c.cpp
mikaelarguedas: surprisingly I can't get to receive the right strings if they exceed ~200 bytes length
have to explicitely pass the size through [rosidl_generator_c__String__assignn](https://github.com/ros2/rosidl/blob/000e26e7db092ea31336c00969f36e958e4924b6/rosidl_generator_c/include/rosidl_generator_c/string_functions.h#L45).
fixed in https://github.com/ros2/system_tests/pull/122/commits/3fca722bd8d55128b1d6fdc2e56f12f7bf8a8def

mikaelarguedas:
I will not spend more time iterating on this, I agree it could be split into 10 files but I'm not sure the value is worth spending time on it now

dirk-thomas:
sgtm.

mikaelarguedas:
rerunning CI before merging:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1892)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1443/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1816/)

dirk-thomas:
@mikaelarguedas Not sure if some of these test failures are related to this PR: http://ci.ros2.org/view/nightly/job/nightly_win_deb/lastCompletedBuild/testReport/

mikaelarguedas:
I'm on it, seems that some failures are from #122 and some from #152 
Thanks for pointing it out

PR #124, opened 2016-06-01
add communication tests for bounded arrays

dirk-thomas:
Connect to #123.

Currently the code generated for bounded arrays by some rosidl generators doesn't even compile, e.g. `rosidl_generator_py`, `rosidl_typesupport_connext_c`, `rosidl_typesupport_opensplice_c/cpp`.

PR #127, opened 2016-06-01
use values that fit in a float for testing

mikaelarguedas:
This avoids language specific approximations/rounding on float variables

### DIFF
--- a/test_communication/test/message_fixtures.hpp
+++ b/test_communication/test/message_fixtures.hpp
@@ -65,8 +65,8 @@ get_messages_primitives()
     msg->bool_value = true;
     msg->byte_value = 255;
     msg->char_value = '\255';
-    msg->float32_value = 1.11f;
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The same will also be necessary for the new bounded array fixtures from #124.

mikaelarguedas:
Indeed, I don't think we should hold this PR waiting for message generation to be fixed (to enable BoundedArray testing). I would be in favour of porting the change to #124 in another PR.

dirk-thomas:
Sounds good to me: +1

dirk-thomas:
I updated #124 to use similar values: a24113d3c662e3454e832735419e9f708d1bbe1d I hope they work fine for your use case.

mikaelarguedas:
Thanks, as long as they can be expressed a sum of power of 2 it should be ok whatever the language.

PR #128, opened 2016-06-02
fix use of bracket

mikaelarguedas:
It seems that https://github.com/ros2/system_tests/commit/4f05870ee0fa897cc9a950be59d3219fc3956ebc while fixing some warnings changed the test behavior.
I assume that the "expected" message would be a message with only empty arrays (std::vectors). But adding a second pair of curly brackets results in initializing all vectors with one element set to its default value (e.g. `msg->bool_values={{}}` will result in `{false}`)
It seems that `{{` is needed only when you want to initialize a vector with more than one element.

This would PR allows to get back to what I think was the expected behavior.

Pending CI jobs to confirm that no new warnings show up:
 * Linux [Build Status](http://ci.ros2.org/job/ci_linux/1377)
 * OSX [Build Status](http://ci.ros2.org/job/ci_osx/1073/)
 * Windows [Build Status](http://ci.ros2.org/job/ci_windows/1382)

### DIFF
--- a/test_communication/test/message_fixtures.hpp
+++ b/test_communication/test/message_fixtures.hpp
@@ -162,20 +162,20 @@ get_messages_dynamic_array_primitives()
   std::vector<test_communication::msg::DynamicArrayPrimitives::SharedPtr> messages;
   {
     auto msg = std::make_shared<test_communication::msg::DynamicArrayPrimitives>();
-    msg->bool_values = {{}};
-    msg->byte_values = {{}};
-    msg->char_values = {{}};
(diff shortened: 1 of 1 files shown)


dirk-thomas:
LGTM

jacquelinekay:
You've probably seen this, but if you're curious to learn more about list initialization syntax:

http://en.cppreference.com/w/cpp/language/list_initialization

jacquelinekay:
Some people think the world should just move on to the next letter of the alphabet https://dlang.org/

PR #129, opened 2016-06-06
use the new exit handler that does not tear down for the publisher

tfoote:
Connects to ros2/launch#35

Requires ros2/launch#35

Leverage new exit handler in pub sub tests


tfoote:
deprecated with the new version of launch. Closing.

PR #130, opened 2016-06-06
add back RCL_ASSERT_RMW_ID_MATCHES for matching implementations

tfoote:
This was removed [here](https://github.com/ros2/system_tests/commit/e37c8391a30d2ca28b88ddb0f14ac85cea52041f) can still be used when testing on the same implementation.


dirk-thomas:
This seems to be a duplicate of #126 but only cover part of the cases.

tfoote:
Yes, duplicate

PR #131, opened 2016-06-06
wait a bit longer on the executor notification test

codebot:
It's always a bit unsettling when "just waiting a bit longer" seems to "fix" something, but this seems to eliminate a flaky test failure on Windows that I can reproduce locally only when running the test in the Command Prompt (cmd.exe).

I thought that Jenkins would be running the test binaries through some sort of better i/o capture mechanism, and when I run the binaries using ConEmu (which I thought would be more similar to what Jenkins does) I can't reproduce the failures until I take the timeout down to 3 milliseconds. But anyway making this timeout a bit longer (50 milliseconds) seems to "fix" the test on Command Prompt (I've run it many hundred times now), so I think this is an improvement. :chicken:

### DIFF
--- a/test_rclcpp/test/test_executor.cpp
+++ b/test_rclcpp/test/test_executor.cpp
@@ -148,7 +148,7 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), notify) {
       timer_promise.set_value();
       timer.cancel();
     });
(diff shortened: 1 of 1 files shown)


wjwwood:
Since this isn't a performance test, I think it's fine to increase the period of time that we will wait for the timer future. Also, it's not like we're slowing down all runs of the test (as a `sleep(N)` and then try again style fix would), just giving slower systems more room to get the right answer.

+1

PR #132, opened 2016-06-09
Use wait_for_service to make Service tests less flaky

wjwwood:
Connects to ros2/ros2#215


test_communication/test/test_requester.cpp
```diff
@@ -35,14 +36,16 @@ int request(
   size_t number_of_cycles = 5)
 {
   int rc = 0;
-  auto start = std::chrono::steady_clock::now();
-
   auto requester = node->create_client<T>(std::string("test_service_") + service_type);
+  if (!requester->wait_for_service(30_s)) {
```
dirk-thomas: Currently the CTest timeout is 15s (https://github.com/ros2/system_tests/blob/50669368f15265ba3ca1e0c78a8a9086d8247c75/test_communication/CMakeLists.txt#L156). So the 30s in the code don't make much sense and might mislead about what is happening.

Same below.

wjwwood: Good point, I think I addressed all the instances in 3bb5fda.

dirk-thomas:
+1

jacquelinekay:
My only comment is: wow, you weren't kidding, these maximum wait times are crazy long.

test_rclcpp/test/test_client_wait_for_service_shutdown.cpp
```diff
@@ -0,0 +1,51 @@
...
+  auto client = node->create_client<test_rclcpp::srv::AddTwoInts>("wait_for_service_shutdown");
+
+  auto shutdown_thread = std::thread([]() {
+    std::this_thread::sleep_for(1_s);
+    rclcpp::shutdown();
+  });
+  auto start = std::chrono::steady_clock::now();
+  client->wait_for_service(15_s);
+  auto end = std::chrono::steady_clock::now();
+  ASSERT_LE(end - start, 10_s);
```
jacquelinekay: So we potentially have to wait 9 seconds after calling shutdown for wait_for_service to exit?

wjwwood: Normally I would expect it to take exactly 1 second. I'll just tolerate up to 9 secs to avoid flaky failures. I didn't see any reason to make a small margin.

jacquelinekay: Yeah... I guess this is more about testing usability than testing correctness. If it actually takes ~9 seconds for this thread to respond to shutdown most of the time, then something is really, really wrong. But if it's an anomalous event that randomly occurs on our buildfarm due to, I don't know, scheduling contention with system threads, I guess this large margin is OK.

wjwwood: I agree, but for me the purpose of these tests is to assert we eventually get the right answer. Basically asserting correctness. I think the other thing you're talking about is more like performance testing and is also important to do either automated or as acceptance testing, but it's also inherently flaky in my experience.

jacquelinekay:
+1, I tested the new wait_for_service_shutdown test locally and it's passes in just a second or two.

PR #136, opened 2016-06-15
require CMake 3.5

dirk-thomas:
Connect to ament/ament_cmake#72

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.5)
 
 project(test_communication)
 
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -126,7 +126,7 @@ if(BUILD_TESTING)
 
       # TODO(dirk-thomas) publishing DynamicArrayPrimitives in OpenSplice is buggy
       # https://github.com/ros2/rclcpp/issues/219
-      if(NOT "${TEST_MESSAGE_TYPE}" STREQUAL "DynamicArrayPrimitives" OR NOT "${rmw_implementation1}" STREQUAL "rmw_opensplice_cpp")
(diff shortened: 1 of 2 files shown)


test_communication/CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ if(BUILD_TESTING)
     endforeach()
 
     # TODO different vendors can't talk to each other right now
-    if("${rmw_implementation1} " STREQUAL "${rmw_implementation2} " OR ("${rmw_implementation1} " STREQUAL "rmw_connext_cpp " AND "${rmw_implementation2} " STREQUAL "rmw_connext_dynamic_cpp ") OR ("${rmw_implementation1} " STREQUAL "rmw_connext_dynamic_cpp " AND "${rmw_implementation2} " STREQUAL "rmw_connext_cpp "))
+    if(rmw_implementation1 STREQUAL rmw_implementation2 OR (rmw_implementation1 STREQUAL "rmw_connext_cpp" AND rmw_implementation2 STREQUAL "rmw_connext_dynamic_cpp") OR (rmw_implementation1 STREQUAL "rmw_connext_dynamic_cpp" AND rmw_implementation2 STREQUAL "rmw_connext_cpp"))
```
wjwwood: Can this be made into a multiline if statement to keep the line short now? I can't remember if something changed between 2.8 and 3.5 that makes this possible. I saw some multiline if statements in one of the other pr's.

dirk-thomas: I can apply that change afterwards since the line was already long before.

wjwwood: That's fine, I was just wondering if it was long before because we weren't supposed to do multiline if statements before (if older versions of CMake didn't fully support it or something).

dirk-thomas: I am wondering why the linter isn't complaining about it in the first place (or if this file isn't being linted at all?).

dirk-thomas: `lint_cmake` is actually checking the code and is fine with it...

Anyway I guess 519c7506f28f551f1b15bf1949cb43b5b375bd6b is a more readable version of the condition.

PR #135, opened 2016-06-15
disable fastrtps<->opensplice tests on osx

gerkey:
Waiting on https://github.com/ros2/ros2/issues/211 to be addressed.

* Linux:
 * [Build Status](http://ci.ros2.org/job/ci_linux/1447/)
* OSX:
 * [Build Status](http://ci.ros2.org/job/ci_osx/1130/)
* Windows:
 * [Build Status](http://ci.ros2.org/job/ci_windows/1463/)

PR #139, opened 2016-06-20
disable fastrtps<->opensplice tests on osx (redo of #135)

gerkey:
Redo of #135: same change, but on another branch to avoid whatever I did wrong on the other branch.


dirk-thomas:
This is obsolete with eProsima/ROS-RMW-Fast-RTPS-cpp#40.

PR #144, opened 2016-06-23
update test times

dirk-thomas:
If test cycles start as infrequent as once per second that implies that after discovery finished the average wasted time is .5 seconds. And if multiple different messages in a cycle are then spaced out by 100ms that can add up to another .5 seconds (for five different messages, e.g. `DynamicArrayPrimitives`).

Since with all the matrix tests we run more than 220 tests in the `test_communication` package alone it makes sense to reduce the amount of wasted time per test as much as possible. Therefore this patch changes the timing to start new cycles at least 10 times per second and only use 10ms between different messages within one cycle. That should reduce the average time per test by .45 - .9 seconds on average. While not all tests are updated the majority of them are (`test_publisher_subscriber__*`) which will safe about 2 minutes:

* Before: http://ci.ros2.org/job/ci_linux/1517/ testing took 680s (only using Connext and OpenSplice, hence only 148 tests)
* After: http://ci.ros2.org/job/ci_linux/1520/ testing took only 619s, which is 61s less / 9% faster

### DIFF
--- a/test_communication/test/test_publisher.cpp
+++ b/test_communication/test/test_publisher.cpp
@@ -26,7 +26,7 @@ void publish(
   rclcpp::Node::SharedPtr node,
   const std::string & message_type,
   std::vector<typename T::SharedPtr> messages,
-  size_t number_of_cycles = 5)
+  size_t number_of_cycles = 100)
 {
   auto start = std::chrono::steady_clock::now();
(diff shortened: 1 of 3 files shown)


wjwwood:
seems reasonable to me, +1

PR #145, opened 2016-06-27
use sleep if wait_for_service throws

dirk-thomas:
With this patch Connext also sleeps since it doesn't implement `wait_for_service` yet (rather then failing these service tests).

Before: http://ci.ros2.org/job/ci_linux/1570/
After: http://ci.ros2.org/job/ci_linux/1571/


wjwwood:
I'm currently working on the fix for this, but it will be a few days probably. Please mention this on the ticket which should in do it once merged: https://github.com/ros2/rmw_connext/issues/168

Lgtm

PR #146, opened 2016-06-29
revert test requester timing

dirk-thomas:
Revert / modify the timing changes form #144 for the `test_requester`. Since the variable `wait_between_services` does not do what it name implies but defines the timeout how long to wait for a response to arrive the shorter timing doesn't apply to it.

Before: http://ci.ros2.org/job/ci_linux/1589/testReport/(root)/test_requester_replier_cpp__Primitives__rmw_connext_cpp_/
After http://ci.ros2.org/job/ci_linux/1590/testReport/(root)/test_requester_replier_cpp__Primitives__rmw_connext_cpp_/

Please review.

### DIFF
--- a/test_communication/test/test_requester.cpp
+++ b/test_communication/test/test_requester.cpp
@@ -35,7 +35,7 @@ int request(
       typename T::Response::SharedPtr
     >
   > services,
-  size_t number_of_cycles = 100)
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
lgtm

PR #147, opened 2016-06-30
remove duplicates of test_subscription_valid_data_cpp, fix skipped tests on Windows

dirk-thomas:
The Windows builds have ~150 skipped tests (http://ci.ros2.org/job/ci_windows/1572/testReport/). Simply because the executable is in the wrong location (http://ci.ros2.org/job/ci_windows/1572/testReport/(root)/projectroot/test_publisher_subscriber_cpp__rmw_connext_cpp__Empty/).

Additionally the `test_subscription_valid_data_cpp` test doesn't use any arguments but is being invoked with every message file in the package (http://ci.ros2.org/job/ci_windows/1572/testReport/(root)/projectroot/) which is 19 times per rmw impl.

Build with this patch http://ci.ros2.org/job/ci_windows/1580/testReport/(root)/projectroot/

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -51,7 +51,7 @@ if(BUILD_TESTING)
     SKIP_INSTALL
   )
 
-  function(custom_test target)
+  function(custom_test target with_message_argument)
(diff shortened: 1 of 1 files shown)


wjwwood:
lgtm

PR #148, opened 2016-07-01
remove noops

dirk-thomas:
Since these packages don't export any libraries using these variables doesn't make any sense.

Please review.


mikaelarguedas:
lgtm

PR #149, opened 2016-07-01
add tests for new param helpers

gerkey:
Connects to ros2/rclcpp#233.


dirk-thomas:
@gerkey See #150.

test_rclcpp/test/test_local_parameters.cpp
```diff
@@ -25,6 +25,9 @@
 #ifdef RMW_IMPLEMENTATION
 # define CLASSNAME_(NAME, SUFFIX) NAME ## __ ## SUFFIX
 # define CLASSNAME(NAME, SUFFIX) CLASSNAME_(NAME, SUFFIX)
+# define QUOTE(arg) #arg
+# define STRINGER(arg) QUOTE(arg)
```
dirk-thomas: Since `QUOTE` itself doesn't do what the caller might expect I suggest to follow the same naming scheme as the other macros. Both having the same name while the first one has an additional underscore.

gerkey:
Replaced by #155.

PR #150, opened 2016-07-01
fix warnings on Windows

dirk-thomas:
While Windows warnings and errors have no priority anymore I think we should still use the appropriate types for the check here.

Before: http://ci.ros2.org/job/ci_windows/1589/warnings35Result/new/package.-1822083627/
After: http://ci.ros2.org/job/ci_windows/1593/


gerkey:
Thanks!

Issue #137, opened 2016-06-15
[test_communication] Refactor test macros to generate cross rmw and cross rcl tests


mikaelarguedas:
Refactor the `multi_target` macro to generate tests for the different configurations:
- ~~same rcl same rmw (will replace custom_test)~~
- same rcl but different rmw (what it currently does but only for rclcpp)
- same rmw but different rcl
- different rcl and different rmw

To do so, need to fix python char and test template to match current cpp tests

mikaelarguedas:
I guess we actually want to keep custom_test to keep testing both single process and multi process pub/sub so removing first item

PR #152, opened 2016-07-05
Test cross RCL communication

mikaelarguedas:
Follow up of https://github.com/ros2/system_tests/pull/121#issuecomment-225325770
Connects to #137 
Uses a single test template to generate cross-rmw and cross-rcl communication tests


mikaelarguedas:
need to increase timeout, seems like 15sec is a bit short on Windows sometimes (c.f. http://ci.ros2.org/job/ci_windows/1679/consoleFull#console-section-197 test # 90)

mikaelarguedas:
reviving this:
multi_rcl tests all platforms/all message types

Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1800)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1356/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1724/)

Can be refactored and optimized but at least show that cross rcl communication is working

mikaelarguedas:
tested with all rmw_implementations and both rcls
before: 9 tests failed out of 242 ([CI](http://ci.ros2.org/job/ci_osx/1410/consoleFull#console-section-211))
after: 9 tests failed out of 649 ([CI](http://ci.ros2.org/job/ci_osx/1410/consoleFull#console-section-211))
Only pub/sub configuration failing: `BoundedArrayNested` message with `connext_dynamic_cpp` with `rclpy` / C typesupport, ticketed [here](https://github.com/ros2/rmw_connext/issues/205)

CI with default parameters:
Linux: [Build Status](http://ci.ros2.org/job/ci_linux/1877)
OSX: [Build Status](http://ci.ros2.org/job/ci_osx/1412/)
Windows: [Build Status](http://ci.ros2.org/job/ci_windows/1792/)

dirk-thomas:
Can you compare the test results without and with this patch. I would expect a significantly higher number of tests but the referenced jobs only have a few more than the nightly builds.

mikaelarguedas:
Not sure how accurate the total test count is with the retries,
if you compare the number of tests in the test_communication package only (during the first try):
nightly: 94 tests
this patch: 194 tests

test_communication/CMakeLists.txt
```diff
@@ -139,24 +139,18 @@ if(BUILD_TESTING)
         NOT rmw_implementation1 STREQUAL "rmw_opensplice_cpp"
```
dhood: maybe re-enable these too by the same logic

test_communication/test/subscriber_py.py
```diff
@@ -42,17 +42,20 @@ def listener_cb(msg, received_messages, expected_msgs):
...
+    from rclpy.impl.rmw_implementation_tools import get_rmw_implementations
+    from rclpy.qos import qos_profile_default
 
     message_pkg = 'test_communication'
     module = importlib.import_module(message_pkg + '.msg')
     msg_mod = getattr(module, message_name)
 
-    select_rmw_implementation(rmw_implementation)
+    rmw_implementations = get_rmw_implementations()
+    assert(os.environ['RCLPY_IMPLEMENTATION'] in rmw_implementations)
```
dhood: would it make sense to remove this check (and the corresponding import) as it's handled by `select_rmw_implementation`  with a specific exception (https://github.com/ros2/rclpy/blob/886ae658a3dcc3aafc8e7ceb697161df2faac846/rclpy/rclpy/impl/rmw_implementation_tools.py#L66) ?

mikaelarguedas: Good point this way we will either raise a `KeyError` exception if the env variable doesn't exist or a `InvalidRCLPYImplementation` exception if the content is not valid.
Fixed in https://github.com/ros2/system_tests/pull/152/commits/55d6d75f2684f78d3ae17b508ac1c4bbeba893c8

rcl variables renamed in https://github.com/ros2/system_tests/pull/152/commits/fd14f0149fa71f0890539f259ba87d7381ac4c2c

dhood: thanks - one more place for rcl rename: https://github.com/ros2/system_tests/pull/152/files#diff-02dc43db6f966560e7ee414c298df21eR235

wjwwood:
Looks like it is cleared up already, but yes I think using rcl's when referring to multiple kinds of client libraries is ambiguous. However I don't have a good suggested alternative that is as convenient to say or type. The current state looks on to me.

test_communication/CMakeLists.txt
```diff
@@ -113,126 +113,136 @@ if(BUILD_TESTING)
...
+        (NOT TEST_MESSAGE_TYPE STREQUAL "BoundedArrayPrimitives" AND NOT TEST_MESSAGE_TYPE STREQUAL "DynamicArrayPrimitives") OR
+        NOT rmw_implementation1 STREQUAL "rmw_opensplice_cpp"
+      )
+        # TODO(mikaelarguedas) Simpler way to blacklist specific tests (e.g. regex matching)
+        ament_add_nose_test(test_publisher_subscriber${test_suffix}
+          "${CMAKE_CURRENT_BINARY_DIR}/test_publisher_subscriber${test_suffix}_$<CONFIGURATION>.py"
+          WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py"
+          PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
+          APPEND_ENV PYTHONPATH="${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}:${CMAKE_CURRENT_BINARY_DIR}/../../rclpy"
+          APPEND_ENV AMENT_PREFIX_PATH=${ament_index_build_path}
```
dirk-thomas: Where is the variable `ament_index_build_path` being set?

mikaelarguedas: it is not, that looks like an artifact of a copy/paste from the `rclpy`'s CMakeLists when I first added the python tests to this package

Issue #153, opened 2016-07-07
two clients with the same settings behave differently for fastrtps


dhood:
In https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/48 (`test_client_scope_cpp` failures), the first client received the response but the second client did not. It can be fixed by making a change to the QoS reliability settings of the second client.

The second client in that demo should have the same behaviour as the first. I have made a regression test that currently fails for fastrtps: https://github.com/ros2/system_tests/commit/0b024f50539643f9840303c0a0dc70afe6b26eb2

it can be fixed by setting
```
rmw_qos_profile.reliability = RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT;
```
for the second client.

should I open a PR even though the test is failing? or is this ticket sufficient?

wjwwood:
If you've committed the test to a branch, then you might as well open a pull request so it's more visible. Otherwise, it's not clear why the branch is there and if it can be pruned or not in the future.

dhood:
thanks, I've done that now

dhood:
closed with https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/48#issuecomment-236144061
regression test added in https://github.com/ros2/system_tests/issues/156

PR #156, opened 2016-07-09
Add regression test for different behaviour between first and second client

dhood:
connects to ros2/system_tests#153

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -169,6 +169,16 @@ if(BUILD_TESTING)
     custom_launch_test(test_client_scope_cpp
       test_client_scope_server_cpp test_client_scope_client_cpp
       TIMEOUT 30)
+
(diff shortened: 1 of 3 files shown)


dhood:
[Changes to FastRTPS](https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/issues/48#issuecomment-236144061) mean that these tests now pass.

Jobs with buildfarm defaults 
Linux (connext static + fastrtps): [Build Status](http://ci.ros2.org/job/ci_linux/1732/)
OS X (connext static + fastrtps): [Build Status](http://ci.ros2.org/job/ci_osx/1292/)

also seems fine for opensplice on linux/osx, and for windows without opensplice.
will have to be updated if https://github.com/ros2/system_tests/pull/162 or similar is merged because this still allows wait_for_service to fail https://github.com/ros2/system_tests/pull/156/files#diff-13664d4cfbb054445e2fcdc74d40d707R53

PR #161, opened 2016-09-13
Ascii fixture

mikaelarguedas:
This PR homogenize fixture between for cpp and python tests so that we can run multi rcl tests.
Note that char values are set to valid ASCII values the time we decide how to handle encoding.

Tests pass locally on opensplice and connext for all message types.
CI jobs:
- Linux:    [Build Status](http://ci.ros2.org/job/ci_linux/1718)
- OSX:    [Build Status](http://ci.ros2.org/job/ci_osx/1283)
- Windows:    [Build Status](http://ci.ros2.org/job/ci_windows/1639)

### DIFF
--- a/test_communication/test/message_fixtures.py
+++ b/test_communication/test/message_fixtures.py
@@ -98,8 +98,8 @@ def get_msg_primitives():
     msg.bool_value = True
     msg.byte_value = bytes([1])
     msg.char_value = '\x01'
-    msg.float32_value = float(1.125)
-    msg.float64_value = float(1.125)
+    msg.float32_value = float(1.0)
+    msg.float64_value = float(1.0)
     msg.int8_value = 1
     msg.uint8_value = 1
     msg.int16_value = 1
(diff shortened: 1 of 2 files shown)


test_communication/test/message_fixtures.py
```diff
@@ -98,17 +98,18 @@ def get_msg_primitives():
...
-    msg.int8_value = int(1)
-    msg.uint8_value = int(1)
-    msg.int16_value = int(1)
-    msg.uint16_value = int(1)
-    msg.int32_value = int(1)
-    msg.uint32_value = int(1)
-    msg.int64_value = int(1)
-    msg.uint64_value = int(1)
+    msg.float32_value = float(1.0)
+    msg.float64_value = float(1.0)
```
dirk-thomas: I would rather check for an actual float value.

Isn't C++ using 1.125? https://github.com/ros2/system_tests/pull/161/files#diff-d88127361752475ebfb4ee2872e5bc4bR70

mikaelarguedas: actually no c++ is using 1.0 and only in this specific message (c.f. https://github.com/ros2/system_tests/blob/master/test_communication/test/message_fixtures.hpp#L106) that's why I changed python accordingly. I hesitated which one to change and given that this message seemed to aim to send 1 in every field I modified the python one.

I don't have srtong feeling about it and change both of them for 1.125 if it makes more sense to everyone

dirk-thomas: Thanks for clarifying. This is only one of the four values. And the other ones cover the value range already. Sounds good to me as is then.

PR #162, opened 2016-09-19
remove blocks and workarounds on service tests

wjwwood:
This should not be merged with eProsima/ROS-RMW-Fast-RTPS-cpp#56, but can be used for testing it, if Connext is not tested at the same time.

This can be merged once Connext has been updated to support wait_for_service complete as well (in progress).

Connects to eProsima/ROS-RMW-Fast-RTPS-cpp#56

### DIFF
--- a/test_communication/test/test_requester.cpp
+++ b/test_communication/test/test_requester.cpp
@@ -109,9 +109,6 @@ int main(int argc, char ** argv)
   std::string service = argv[1];
   auto node = rclcpp::Node::make_shared(std::string("test_requester_") + service);
 
-  // NOTE(esteve): sleep for two seconds to let the service start up
-  std::this_thread::sleep_for(std::chrono::seconds(2));
-
   int rc;
   if (service == "Empty") {
     rc = request<test_communication::srv::Empty>(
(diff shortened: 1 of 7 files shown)


dhood:
This might be ready to be removed as well, right? https://github.com/ros2/system_tests/blob/f569a353ea51924e44ed42eb3fb7d33a7277bd84/test_communication/test/test_requester.cpp#L113 (couldn't add a line comment)

wjwwood:
Yeah, but it would need to be replaced with a `wait_for_service` call. I just started CI, but if that's looking ok, I'll try to make this change and rerun them.

dhood:
is this `wait_for_service` call not sufficient? https://github.com/ros2/system_tests/blob/f569a353ea51924e44ed42eb3fb7d33a7277bd84/test_communication/test/test_requester.cpp#L41

wjwwood:
Ah, yeah it is. Sorry, should have read more of the context.

wjwwood:
Removed it in 2bc1097.

wjwwood:
I'm going to close this pr and push the changes to a different branch name, that way I can test the other changes on this branch name without these because they break Connext.

wjwwood:
New branch name is `local_graph_changes` to match the upcoming Connext changes which should make these changes work.

PR #164, opened 2016-09-21
Local graph changes

wjwwood:
Connects to ros2/rmw_connext#168.

### DIFF
--- a/test_rclcpp/test/test_publisher.cpp
+++ b/test_rclcpp/test/test_publisher.cpp
@@ -14,6 +14,7 @@
 
 #include <chrono>
 #include <iostream>
+#include <string>
 
 #include "gtest/gtest.h"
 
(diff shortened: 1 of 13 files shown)


test_rclcpp/test/test_publisher.cpp
```diff
@@ -31,52 +34,43 @@
 
 // Short test for the const reference publish signature.
 TEST(CLASSNAME(test_publisher, RMW_IMPLEMENTATION), publish_with_const_reference) {
-  rclcpp::init(0, nullptr);
-  auto node = rclcpp::Node::make_shared("test_publisher");
-
-  auto publisher = node->create_publisher<test_rclcpp::msg::UInt32>("test_publisher", 10);
-
+  // topic name
```
dirk-thomas: The comment doesn't contain more information then the following variable name.

Same below.

test_rclcpp/test/test_publisher.cpp
```diff
@@ -31,52 +34,43 @@
...
+  std::string topic_name = "test_publish_with_const_reference";
+  // code to create the callback and subscription
```
dirk-thomas: I would suggest to remove the `code to` part in the comment.

Same multiple times below.

test_rclcpp/test/test_publisher.cpp
wjwwood: 33c115d

test_rclcpp/test/test_publisher.cpp
wjwwood: 33c115d

PR #166, opened 2016-10-01
comply with stricter cpplint rules

dirk-thomas:
Update style to pass new cpplint check (ament/ament_lint#59).

http://ci.ros2.org/job/ci_linux/1763/

### DIFF
--- a/test_rclcpp/test/test_services_server.cpp
+++ b/test_rclcpp/test/test_services_server.cpp
@@ -16,6 +16,8 @@
 
 #include <test_rclcpp/srv/add_two_ints.hpp>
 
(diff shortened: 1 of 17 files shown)


test_rclcpp/test/test_services_server.cpp
```diff
@@ -16,6 +16,8 @@
 
 #include <test_rclcpp/srv/add_two_ints.hpp>
 
+#include <memory>
```
dirk-thomas: The same question as https://github.com/ros2/demos/pull/80#discussion_r81456795 applies here.

PR #177, opened 2016-11-24
Mark blacklisted tests as skipped

mikaelarguedas:
Use the new `SKIP_TEST` argument
connects to https://github.com/ros2/ros2/issues/290
depends on https://github.com/ament/ament_cmake/pull/80

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -172,19 +172,19 @@ if(BUILD_TESTING)
       )
         set(SKIP_TEST "SKIP_TEST")
       endif()
-        # TODO(mikaelarguedas) Simpler way to blacklist specific tests (e.g. regex matching)
-        ament_add_nose_test(test_publisher_subscriber${test_suffix}
-          "${CMAKE_CURRENT_BINARY_DIR}/test_publisher_subscriber${test_suffix}_$<CONFIGURATION>.py"
-          WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py"
-          PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
-          APPEND_ENV PYTHONPATH="${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}:${CMAKE_CURRENT_BINARY_DIR}/../../rclpy"
-          APPEND_LIBRARY_DIRS "${_append_library_dirs}"
-          TIMEOUT 15
-          ${SKIP_TEST})
-        set_tests_properties(
-          test_publisher_subscriber${test_suffix}
-          PROPERTIES DEPENDS "test_publisher_cpp__${rmw_implementation1};test_subscriber_cpp__${rmw_implementation2}"
-        )
+      # TODO(mikaelarguedas) Simpler way to blacklist specific tests (e.g. regex matching)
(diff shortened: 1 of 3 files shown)


test_communication/CMakeLists.txt
```diff
@@ -68,7 +68,16 @@ if(BUILD_TESTING)
...
-        if((NOT TEST_MESSAGE_TYPE STREQUAL "BoundedArrayPrimitives" AND NOT TEST_MESSAGE_TYPE STREQUAL "DynamicArrayPrimitives") OR NOT rmw_implementation STREQUAL "rmw_opensplice_cpp")
+        if(
+          (TEST_MESSAGE_TYPE STREQUAL "BoundedArrayPrimitives" OR TEST_MESSAGE_TYPE STREQUAL "DynamicArrayPrimitives") AND
+          rmw_implementation STREQUAL "rmw_opensplice_cpp"
+        )
+          ament_add_test(
+            "${target}${target_suffix}__${TEST_MESSAGE_TYPE}"
+            COMMAND "$<TARGET_FILE:${target}${target_suffix}>" "${TEST_MESSAGE_TYPE}"
+            SKIP_TEST
+          )
```
dirk-thomas: Instead of duplicating the function call with all its arguments I would suggest to use a variable to either pass or not pass `SKIP_TEST`.

mikaelarguedas: fixed in https://github.com/ros2/system_tests/pull/177/commits/c9d1695b7d6e96b9110383cee708eac6566239d0

test_communication/CMakeLists.txt
```diff
@@ -65,21 +65,27 @@ if(BUILD_TESTING)
     if(with_message_argument)
       # adding test for each message type
       foreach(message_file ${message_files})
+        set(_SKIP_TEST "")
```
dirk-thomas: Since the variable has no local scope why should it be prefixed with an underscore? No other variable in this file is.

mikaelarguedas: It was to avoid confusion with the value stored in it that is also `SKIP_TEST` but the leading underscore is misleading indeed, I'll find a better name

mikaelarguedas: updated https://github.com/ros2/system_tests/pull/177/commits/c9a9a8715ac9508ff3b62f0d6c713c6a359580f2

test_communication/CMakeLists.txt
```diff
@@ -166,18 +163,28 @@ if(BUILD_TESTING)
...
+          WORKING_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py"
+          PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
+          APPEND_ENV PYTHONPATH="${CMAKE_CURRENT_BINARY_DIR}/rosidl_generator_py:${CMAKE_CURRENT_SOURCE_DIR}:${CMAKE_CURRENT_BINARY_DIR}/../../rclpy"
+          APPEND_LIBRARY_DIRS "${_append_library_dirs}"
+          TIMEOUT 15
+          ${SKIP_TEST})
+        set_tests_properties(
+          test_publisher_subscriber${test_suffix}
+          PROPERTIES DEPENDS "test_publisher_cpp__${rmw_implementation1};test_subscriber_cpp__${rmw_implementation2}"
+        )
```
mikaelarguedas: thanks for pointing it out https://github.com/ros2/system_tests/pull/177/commits/010911cba6f1db9ce893284fd40d08b00b442e64

PR #189, opened 2017-02-10
Use -Wpedantic

mikaelarguedas:
Connects to https://github.com/ros2/ros2/issues/305

### DIFF
--- a/test_communication/test/test_messages_c.cpp
+++ b/test_communication/test/test_messages_c.cpp
@@ -129,26 +129,28 @@ public:
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
     EXPECT_EQ(0, strcmp(rcl_subscription_get_topic_name(&subscription), topic_name));
 
-    rcl_wait_set_t wait_set_connection = rcl_get_zero_initialized_wait_set();
-    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
-    ret = rcl_wait_set_init(
-      &wait_set_connection,
-      0,  // number_of_subscriptions
-      1,  // number_of_guard_conditions
-      0,  // number_of_timers
-      0,  // number_of_clients
-      0,  // number_of_services
-      rcl_get_default_allocator());
-    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
-    ret = rcl_wait_set_clear_guard_conditions(&wait_set_connection);
-    EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
-    const rcl_guard_condition_t * graph_guard_condition =
-      rcl_node_get_graph_guard_condition(this->node_ptr);
-    ret = rcl_wait_set_add_guard_condition(
-      &wait_set_connection, graph_guard_condition);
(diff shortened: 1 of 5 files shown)


test_communication/test/test_subscriber.cpp
```diff
@@ -57,8 +57,8 @@ rclcpp::subscription::SubscriptionBase::SharedPtr subscribe(
       }
 
       // shutdown node when all expected messages have been received
-      for (auto received : received_messages) {
-        if (!received) {
+      for (auto received_msg : received_messages) {
+        if (!received_msg) {
```
wjwwood: Was this a warning?

test_communication/test/test_messages_c.cpp
```diff
@@ -174,7 +174,7 @@ class CLASSNAME (TestMessagesFixture, RMW_IMPLEMENTATION) : public ::testing::Te
           fini_message(&message);
         });
 
-        rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+        wait_set = rcl_get_zero_initialized_wait_set();
```
wjwwood: Would it be better/possible to narrow the scope above to avoid the shadowing (I assume that was the issue here)?

test_communication/test/test_subscriber.cpp
mikaelarguedas: Actually I think that was not a pedantic warning but a windows warning when I was trying more permissive flags. I think it's because it is shadowing the variable of the same name defined [line 39](https://github.com/ros2/system_tests/blob/087c3743e06eae7b1d82c175835a82d15fe31ab2/test_communication/test/test_subscriber.cpp#L39)

wjwwood: Sounds good to me.

test_communication/test/test_messages_c.cpp
mikaelarguedas: very good point. fixed in https://github.com/ros2/system_tests/pull/189/commits/d444705ab1fab805ec6715b0784e9291f7fcd363

test_communication/test/test_messages_c.cpp
```diff
@@ -129,25 +129,25 @@ class CLASSNAME (TestMessagesFixture, RMW_IMPLEMENTATION) : public ::testing::Te
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string_safe();
     EXPECT_EQ(0, strcmp(rcl_subscription_get_topic_name(&subscription), topic_name));
 
-    rcl_wait_set_t wait_set = rcl_get_zero_initialized_wait_set();
+    rcl_wait_set_t wait_set_connection = rcl_get_zero_initialized_wait_set();
```
dirk-thomas: Why adding `_connection` to this variable? How is the wait set related to a "connection"?

Instead the block could be wrapped in a local scope to avoid the colliding variable name.

mikaelarguedas: You're right it's better to just wrap it in a local scope, which is what @wjwwood suggested initially I think.

Fixed in https://github.com/ros2/system_tests/pull/189/commits/ffce07c9c26559f538a4cda614e2d9cb2cd389a0

PR #190, opened 2017-03-10
Install msgs and fixtures for use by other packages

gerkey:
When implementing tests for `rostopic_echo_py` (https://github.com/ros2/cli_tools/pull/5), I had need of a bunch of messages to push through the YAML and CSV converters. This PR installs the messages and related fixtures so that they can be used in testing other packages.

http://ci.ros2.org/job/ci_linux/2314/
http://ci.ros2.org/job/ci_osx/1769/
http://ci.ros2.org/job/ci_windows/2343/

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -14,13 +14,6 @@ find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_auto REQUIRED)
 ament_auto_find_build_dependencies()
 
-# get the rmw implementations ahead of time
-find_package(rmw_implementation_cmake REQUIRED)
-get_available_rmw_implementations(rmw_implementations2)
-foreach(rmw_implementation ${rmw_implementations2})
-  find_package("${rmw_implementation}" REQUIRED)
-endforeach()
-
 set(message_files
   "msg/BoundedArrayNested.msg"
   "msg/BoundedArrayPrimitives.msg"
@@ -51,6 +44,13 @@ rosidl_generate_interfaces(${PROJECT_NAME}
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
+
+  # get the rmw implementations ahead of time
+  find_package(rmw_implementation_cmake REQUIRED)
+  get_available_rmw_implementations(rmw_implementations2)
+  foreach(rmw_implementation ${rmw_implementations2})
+    find_package("${rmw_implementation}" REQUIRED)
+  endforeach()
 
   function(custom_test target with_message_argument)
     if(with_message_argument)
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -11,45 +11,46 @@ else()
 endif()
 
 find_package(ament_cmake REQUIRED)
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
+
+# get the rmw implementations ahead of time
+find_package(rmw_implementation_cmake REQUIRED)
+get_available_rmw_implementations(rmw_implementations2)
+foreach(rmw_implementation ${rmw_implementations2})
+  find_package("${rmw_implementation}" REQUIRED)
+endforeach()
+
+set(message_files
+  "msg/BoundedArrayNested.msg"
+  "msg/BoundedArrayPrimitives.msg"
+  "msg/Builtins.msg"
+  "msg/DynamicArrayNested.msg"
+  "msg/DynamicArrayPrimitives.msg"
+  "msg/Empty.msg"
+  "msg/Nested.msg"
+  "msg/Primitives.msg"
+  "msg/StaticArrayNested.msg"
+  "msg/StaticArrayPrimitives.msg"
+)
+set(other_message_files
+  "msg/FieldsWithSameType.msg"
+  "msg/UInt32.msg"
+)
+set(service_files
+  "srv/Empty.srv"
+  "srv/Primitives.srv"
+)
+rosidl_generate_interfaces(${PROJECT_NAME}
+  ${message_files}
+  ${other_message_files}
+  ${service_files}
+  DEPENDENCIES builtin_interfaces
+)
 
 if(BUILD_TESTING)
   find_package(ament_lint_auto REQUIRED)
   ament_lint_auto_find_test_dependencies()
-
-  # get the rmw implementations ahead of time
-  find_package(rmw_implementation_cmake REQUIRED)
-  get_available_rmw_implementations(rmw_implementations2)
-  foreach(rmw_implementation ${rmw_implementations2})
-    find_package("${rmw_implementation}" REQUIRED)
-  endforeach()
-
-  set(message_files
-    "msg/BoundedArrayNested.msg"
-    "msg/BoundedArrayPrimitives.msg"
-    "msg/Builtins.msg"
-    "msg/DynamicArrayNested.msg"
-    "msg/DynamicArrayPrimitives.msg"
-    "msg/Empty.msg"
-    "msg/Nested.msg"
-    "msg/Primitives.msg"
-    "msg/StaticArrayNested.msg"
-    "msg/StaticArrayPrimitives.msg"
-  )
-  set(other_message_files
-    "msg/FieldsWithSameType.msg"
-    "msg/UInt32.msg"
-  )
-  set(service_files
-    "srv/Empty.srv"
-    "srv/Primitives.srv"
-  )
-  rosidl_generate_interfaces(${PROJECT_NAME}
-    ${message_files}
-    ${other_message_files}
-    ${service_files}
-    DEPENDENCIES builtin_interfaces
-    SKIP_INSTALL
-  )
 
   function(custom_test target with_message_argument)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
These packages currently only contain tests. From a packaging point of view they don't even have to be released.

Therefore I don't think other packages should start depending on this package. Instead the messages should be extracted into a separate package which the cli tools as well as the tests can then depend on.

test_communication/package.xml
```diff
@@ -11,21 +11,23 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>ament_cmake_auto</build_depend>
+  <build_depend>builtin_interfaces</build_depend>
+  <build_depend>rclcpp</build_depend>
+  <build_depend>rclpy</build_depend>
+  <build_depend>rcl</build_depend>
```
wjwwood: (forgot to actually submit this comment with the review, sorry)

Do these (`rcl*`) packages need to be build depends? I think they could stay as test depends.

gerkey: I think that they should be build depends, so that the call to `ament_auto_find_build_dependencies()` will have the side effect of causing the subsequent call to `rosidl_generate_interfaces()` generating for all the languages.

But I could be very wrong; I've honestly never understood how to use the various `_depend` tags or how they interact with what happens in CMake :)

mikaelarguedas:
I don't see any blocker for doing it.
There are 2 points we should agree on before moving on:
 - Where should this package land? I think that rosidl should work
 - Should this package be IDL only or also provide fixtures for both C, C++ and Python?

Once this package created we should update all packages generating messages and services for testing to use those instead.
I was seeing this work as a follow up refactor rather than holding https://github.com/ros2/cli_tools/pull/5 on it. But I'm fine either way

test_communication/package.xml
mikaelarguedas: I think that @wjwwood is right. You need the generators (rosidl_default_generators) to build the messages but not the client libraries. The client libraries can be used to interact with messages but are not required to generate/build them.

dirk-thomas: The generated type support triggered by `rosidl_generate_interfaces` doesn't depend on the client libraries `rcl*`. The only build dependency necessary for that is `rosidl_default_generators`.

Also `rosidl_default_runtime` should be a run dependency.

wjwwood:
Ok, I'm fine with doing this for now and planning to split it out into a different package then update everything else to use it. However, it would be nice to see that task on the next sprint if possible. I'd even be willing to take on the task, but I fear if we don't schedule it, then it won't get done.

> Where should this package land? I think that rosidl should work

If you're talking about repositories, then yeah `rosidl` can work since all of the other generators that are currently explicitly depended on live there too.

> Should this package be IDL only or also provide fixtures for both C, C++ and Python?

Are you talking about some extra code that helps you run tests over all the messages? I don't see a problem with that since the package will probably be called `test_interfaces` or something like that, then it's clear it's for testing purposes and so having extra test utilities live there sounds fine to me too.

dirk-thomas:
> > Where should this package land? I think that rosidl should work

> If you're talking about repositories, then yeah rosidl can work since all of the other generators that are currently explicitly depended on live there too.

It can't be in `rosidl`. Every message package depends on `rosidl_default_generators` and `rosidl_default_runtime` which are in `rosidl_typesupport`. So putting them there would result in a circular dependency in repository level.

I would suggest either `common_interfaces` or a new repo.

wjwwood:
That's true. `common_interfaces` sounds like a better idea. I'm also ok with a new repository.

mikaelarguedas:
> > > Where should this package land? I think that rosidl should work

> > If you're talking about repositories, then yeah rosidl can work since all of the other generators that are currently explicitly depended on live there too.

> It can't be in rosidl. Every message package depends on rosidl_default_generators and rosidl_default_runtime which are in rosidl_typesupport. So putting them there would result in a circular dependency in repository level.
> I would suggest either common_interfaces or a new repo.

Good point, I'm fine with either of those options. The former avoiding to have a new repo just for that purpose, the latter allowing to keep `common_interfaces` closer to its ROS1 counterpart `common_msgs`.

>  However, it would be nice to see that task on the next sprint if possible.

Agreed, we talked about it for a while and never got around doing it. I don't mind taking it on either, it shouldn't take long.

test_communication/package.xml
gerkey: a2e014c reshuffles the depend tags as suggested, and the desired code generation still works. I learn something new every day.

test_communication/package.xml
```diff
@@ -11,21 +11,24 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>ament_cmake_auto</build_depend>
+  <build_depend>builtin_interfaces</build_depend>
+  <build_depend>rmw_implementation</build_depend>
+  <build_depend>rmw_implementation_cmake</build_depend>
```
dirk-thomas: The `rmw_implementation` packages are only used for the tests and should therefore still be test dependencies.

gerkey: da226ac

gerkey:
I ticketed myself to create the separate repo to hold the reusable stuff: #191.

test_communication/package.xml
```diff
@@ -11,21 +11,24 @@
...
   <test_depend>ament_cmake_nose</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
-  <test_depend>builtin_interfaces</test_depend>
   <test_depend>launch</test_depend>
-  <test_depend>rcl</test_depend>
   <test_depend>rclcpp</test_depend>
   <test_depend>rclpy</test_depend>
+  <test_depend>rcl</test_depend>
```
dirk-thomas: This is now in wrong order. It was correct before being moved.

test_communication/package.xml
```diff
@@ -11,21 +11,24 @@
   <maintainer email="dthomas@osrfoundation.org">Dirk Thomas</maintainer>
   <license>Apache License 2.0</license>
 
+  <build_depend>ament_cmake_auto</build_depend>
+  <build_depend>builtin_interfaces</build_depend>
+  <build_depend>rosidl_default_generators</build_depend>
+
   <buildtool_depend>ament_cmake</buildtool_depend>
```
dirk-thomas: This dependency as well as the `find_package(ament_cmake)` call is redundant now.

test_communication/CMakeLists.txt
```diff
@@ -322,5 +323,7 @@ if(BUILD_TESTING)
   call_for_each_rmw_implementation(targets)
 endif()  # BUILD_TESTING
 
-# TODO should not install anything
+install(FILES test/__init__.py test/message_fixtures.py
+  DESTINATION "${PYTHON_INSTALL_DIR}/${PROJECT_NAME}")
```
dirk-thomas: What about the C++ fixtures?

mikaelarguedas: they were not needed by https://github.com/ros2/cli_tools/pull/5 that motivated this PR in the first place. Fixtures for all languages will be moved in (and installed by) the package that will be created in https://github.com/ros2/system_tests/issues/191

test_communication/package.xml
gerkey: Good to know. But why?

dirk-thomas: Because `ament_cmake_auto` depends on `ament_cmake` and exports that dependency (https://github.com/ament/ament_cmake/blob/7b31d859f3d42b9eb6c84db807d95a5604a6881b/ament_cmake_auto/ament_cmake_auto-extras.cmake#L17).

gerkey: Then should I declare `ament_cmake_auto` is a `build_depend` or a `buildtool_depend` and why?

dirk-thomas: The package should then also call `ament_auto_package` (https://github.com/ament/ament_cmake/blob/7b31d859f3d42b9eb6c84db807d95a5604a6881b/ament_cmake_auto/cmake/ament_auto_package.cmake#L31) instead of `ament_package`.

dirk-thomas: A `buildtool_depend` dependency because you need the dependency for the host system where you are performing the build on (in case of a cross compilation).

gerkey: What difference in behavior should I expect by calling `ament_auto_package()`? I'm not doubting the advice, but trying to understand why each thing is used in a different context. That's especially difficult to determine for these kinds of  changes to the `package.xml` and `CMakeLists.txt` because everything seems to work fine with a wide variety of different combinations and I don't have the context to identify potential problems by inspection.

dirk-thomas: `ament_cmake_auto` aims to automate several steps in your CMake and you need to call its function in order to actually perform all those: see https://github.com/ament/ament_cmake/blob/master/ament_cmake_auto/cmake/ament_auto_package.cmake You either don't need any of these (than you probably want to use `ament_cmake` instead) or you haven't noticed that they are not happening.

gerkey:
ros2/system_tests#192 fixes the minor dependency and cmake usage items that @dirk-thomas raised.

PR #193, opened 2017-03-16
tests for get_parameter_or and set_parameter_if_not_set

wjwwood:
Connects to ros2/rclcpp#309

### DIFF
--- a/test_rclcpp/test/test_local_parameters.cpp
+++ b/test_rclcpp/test/test_local_parameters.cpp
@@ -309,6 +309,73 @@ TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), get_from_node_variant
   EXPECT_EQ(true, got_param);
 }
 
+TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), get_parameter_or) {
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
@wjwwood ping just to make sure this is still in progress and has not been forgotten when merging https://github.com/ros2/rclcpp/issues/309

wjwwood:
Oops, good catch 😛

mikaelarguedas:
@wjwwood FYI: this PR introduced [new warnings](http://ci.ros2.org/view/nightly/job/nightly_win_deb/420/) on MSVC. It's due to ParameterVariant defining integer parameters as int64 and these tests use `int`

wjwwood:
@mikaelarguedas shoot, I missed that (symptom of not having green jobs normally I guess). I also didn't notice because it was introduced before I started tracking these as the Build cop, so I file the warning as "existing" and to look into it later.

I'll try to avoid that in the future.

wjwwood:
That's frustrating, I could use a concept with a constraint to limit the possible data types for these functions, but I'm not sure those are ready for prime time:

http://en.cppreference.com/w/cpp/language/constraints

I'll open a pr soon to fix it.

PR #198, opened 2017-04-02
Add missing exec dep on builtin_interfaces

mikaelarguedas:
`test_communication` fixtures need `builtin_interfaces`. Unveiled by `cli_tools/test_primitives` using these for testing.
This should fix the failing `cli_tools/test_primitives` test on all nightlies: e.g http://ci.ros2.org/view/nightly/job/nightly_linux_release/lastSuccessfulBuild/testReport/(root)/test_primitives/test_primitives/

### DIFF
--- a/test_communication/package.xml
+++ b/test_communication/package.xml
@@ -16,8 +16,8 @@
 
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
 
-  <exec_depend>rosidl_default_runtime</exec_depend>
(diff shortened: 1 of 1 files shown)


dirk-thomas:
I don't think `test_communication` should export its dependencies. The rational why it didn't before is that it is not intended that other package depend in it. In the future this package does not need to be released since it only contains tests. Therefore it can't be depended on by other packages.

PR #200, opened 2017-04-11
Fix tests for many core machines

gerkey:
This PR addresses failures that have come up since we started running on 96-core ARM servers:

* In general, cap the internal test parallelism at 16, independent of how many cores the machine has.
* In one case, adjust the spin_until logic to wait for *all* futures to complete instead of waiting for just one (the old behavior is presumably reliable for a low level of parallelism, but it's not when you go wide).
* In one case, adjust the subscriber-side queue length to accommodate the expected number of messages (again, the old behavior is presumably reliable for low parallelism).

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
 #include <chrono>
 #include <limits>
 #include <memory>
@@ -64,7 +65,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
 
   rclcpp::executors::MultiThreadedExecutor executor;
   // Try to saturate the MultithreadedExecutor's thread pool with subscriptions
-  for (uint32_t i = 0; i < executor.get_number_of_threads(); i++) {
+  for (uint32_t i = 0; i < std::min<unsigned int>(executor.get_number_of_threads(), 16); i++) {
     auto sub = node->create_subscription<test_rclcpp::msg::UInt32>(node_topic_name, 16, callback,
         callback_group);
     subscriptions.push_back(sub);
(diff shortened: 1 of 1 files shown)


gerkey:
* http://ci.ros2.org/job/ci_osx/1885/
* http://ci.ros2.org/job/ci_windows/2559/
* http://ci.ros2.org/job/ci_linux-aarch64/59/
* http://ci.ros2.org/job/ci_linux/2438/

gerkey:
I agree that it should work for a higher number of cores, and from what I saw we could get to ~50 with the current test configuration. Going higher would have (I believe) required modifying the heuristic timeout logic, which I wanted to avoid, on the assumption that the current logic has been iteratively tuned to at least handle the 16-core case. Plus, now that we're running multiple Jenkins executors per server, we should avoid using all the available cores, as we'll inevitably one day have two jobs try to do that at the same time.

gerkey:
CI shows that the targeted test failure was fixed. Various other failures are shown, but they do not appear to be related to this PR.

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -202,10 +203,11 @@ TEST(CLASSNAME(test_multithreaded, RMW_IMPLEMENTATION), multi_consumer_clients)
       }
       results.push_back(pair.first->async_send_request(pair.second));
     }
-    // Wait on the future produced by the first request
-    auto result = executor.spin_until_future_complete(results.back());
-
-    ASSERT_EQ(result, rclcpp::executor::FutureReturnCode::SUCCESS);
+    // Wait on each future
+    for (uint32_t i = 0; i < results.size(); i++) {
```
dirk-thomas: I would assume that not using `size_t` for the type would result in a compiler warning on some platforms.

dirk-thomas: Nitpick: we use `++i`.

gerkey: #201

gerkey: #201

PR #201, opened 2017-04-12
fix type and style

gerkey:
Address issues lingering from #200 .

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -65,7 +65,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
 
   rclcpp::executors::MultiThreadedExecutor executor;
   // Try to saturate the MultithreadedExecutor's thread pool with subscriptions
-  for (uint32_t i = 0; i < std::min<unsigned int>(executor.get_number_of_threads(), 16); i++) {
+  for (uint32_t i = 0; i < std::min<size_t>(executor.get_number_of_threads(), 16); ++i) {
     auto sub = node->create_subscription<test_rclcpp::msg::UInt32>(node_topic_name, 16, callback,
         callback_group);
     subscriptions.push_back(sub);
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -297,7 +297,7 @@ static inline void multi_access_publisher(bool intra_process)
     };
   std::vector<rclcpp::timer::TimerBase::SharedPtr> timers;
   // timers will fire simultaneously in each thread
-  for (uint32_t i = 0; i < std::min<unsigned int>(executor.get_number_of_threads(), 16); i++) {
+  for (uint32_t i = 0; i < std::min<size_t>(executor.get_number_of_threads(), 16); i++) {
```
dirk-thomas: Nitpick: `++i`

Multiple times in this file.

gerkey: Ah, there was pre-existing non-conformant style.

Can the linters be made to check for that? It's not good a use of humans' time to check such things...

f060cd1

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -204,13 +204,13 @@ TEST(CLASSNAME(test_multithreaded, RMW_IMPLEMENTATION), multi_consumer_clients)
       results.push_back(pair.first->async_send_request(pair.second));
     }
     // Wait on each future
-    for (uint32_t i = 0; i < results.size(); i++) {
+    for (uint32_t i = 0; i < results.size(); ++i) {
```
dirk-thomas: Doesn't this line still generate a compiler warning?

Same for all the other loops declaring `i` as `uint32_t`.

wjwwood: I don't think so, because it is a `<` comparison it's fine for `i` to be a different size from `results.size()` since they're both unsigned. That's my guess at least. Anyways `i` needs to be `uint32_t` so it can be assigned into the message field below. Once we sort out the farm I'll rerun CI on this branch to see for sure.

wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/2449/)
- Linux-aarch64:
  - [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/74/)
- macOS:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1895/) (1 known warning)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/2568/) (know flaky tests, no warnings)

PR #206, opened 2017-06-07
add test for avoid_ros_namespace_conventions qos

wjwwood:
New test based on @dhood's suggestion. This should be enough to put the rest of the connected pr's in review without waiting to refactor topic names and types.

Connects to ros2/rmw#106


wjwwood:
I could have added a test that has only one side drop the prefix and then ensure they never talk, but I couldn't find a good way to do that with it being just flakey and reasonably fast.

In the future I can augment this with topic names and types has a "raw" mode or whatever.

PR #208, opened 2017-06-15
add option for security tests

mikaelarguedas:
This creates and run the security tests only if `-DSECURITY=ON` is passed to cmake


mikaelarguedas:
For reference, I kept the value OFF by default to match Fast-RTPS's default https://github.com/eProsima/Fast-RTPS/blob/e6a75ccaaa8cf0eba3cae60b64750a8fa24db2e8/CMakeLists.txt#L208

PR #216, opened 2017-08-10
testing array longers than 101

mikaelarguedas:
This is mostly to make sure that we don't rely on the magic 101 [here](https://github.com/ros2/rmw_fastrtps/blob/8623ccde0fbdeab8142ed4062978c977446327f0/rmw_fastrtps_cpp/include/rmw_fastrtps_cpp/TypeSupport_impl.hpp#L772) or any other arbitrary limitation on arrays

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2976/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/415/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3085/)

### DIFF
--- a/test_communication/test/message_fixtures.hpp
+++ b/test_communication/test/message_fixtures.hpp
@@ -236,7 +236,7 @@ get_messages_dynamic_array_primitives()
   {
     auto msg = std::make_shared<test_communication::msg::DynamicArrayPrimitives>();
     // check sequences with more then 100 elements
-    const size_t size = 101;
+    const size_t size = 1000;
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
@allenh1 is currently looking at https://github.com/ros2/rmw_fastrtps/issues/72.
Now that the fixtures here don't assume 257 for max string length and 101 for max array size we should be able to remove both from rmw_fastrtps.

PR #218, opened 2017-08-10
use unbuffered Python in launch files

dirk-thomas:
This ensures that the output of the Python executables is visible when something goes south. That avoids speculations around why a test might have failed (see `test_publisher_subscriber__Nested__rclpy__rmw_fastrtps_cpp_.test_publisher_subscriber` in ros2/build_cop#41).

I don't plan to run CI for this simple change.

### DIFF
--- a/test_communication/test/test_publisher_subscriber.py.in
+++ b/test_communication/test/test_publisher_subscriber.py.in
@@ -13,12 +13,16 @@ def test_publisher_subscriber():
     publisher_cmd = ['@TEST_PUBLISHER_EXECUTABLE@', '@TEST_MESSAGE_TYPE@']
     subscriber_cmd = ['@TEST_SUBSCRIBER_EXECUTABLE@', '@TEST_MESSAGE_TYPE@']
 
+    publisher_env = dict(os.environ)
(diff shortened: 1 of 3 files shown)


dhood:
these scripts are being added via `ament_add_nose_test`, so I would have thought that https://github.com/ament/ament_cmake/pull/98/files#diff-29a991acf62a8fc4460ed46cdb27c46eR88 would already take care of this. Is there a difference?

mikaelarguedas:
My understanding is that the high level interpreter gets the `-u` but given that these process started through launch use [another interpreter](https://github.com/ros2/system_tests/pull/218/files#diff-c271176388ec27db9b52f6da576eac72R23) they need to be passed the unbuffered request in their respective environments

dhood:
ah I didn't see the nesting of python, that makes sense, thank you

mikaelarguedas:
setting it in the environment of the caller would solve it though because that environment is copied, modified and passed to the underlying interpreters.

dirk-thomas:
Setting it in the environment when registering the test wouldn't apply if the user runs the launch file from the console. Therefore I think this is the better location.

Issue #191, opened 2017-03-10
Move messages and fixtures from test_communication into another repo


gerkey:
ros2/system_tests#190 arranges for installation of messages and helpful Python code that's now being used elsewhere (ros2/cli_tools#5). But it would be better to move that stuff into a different repo (https://github.com/ros2/system_tests/pull/190#issuecomment-285801282).

mikaelarguedas:
see https://github.com/ros2/rcl_interfaces/pull/16#issue-219022063 for what remains to be done

mikaelarguedas:
@gerkey is this still on your todo list ?

gerkey:
It's certainly not near the top of my stack. If it becomes urgent, I'm happy to work on it.

mikaelarguedas:
Should we unassign this until someone that has time picks it up ?

dirk-thomas:
We have already copied the messages / fixtures to a second location. Therefore in my opinion we should follow up on also using them from here and not build up more unnecessary technical debt.

mikaelarguedas:
> Therefore in my opinion we should follow up on also using them from here

:+1: this will allow to reduce the amount of code to maintain and build time to not have to build messages for every package needing to test anything message related, so I also think that we should follow-up on this sooner than later.

dirk-thomas:
@gerkey We assigned this ticket two weeks ago for the sprint. Is there any update on this?

mikaelarguedas:
* Linux [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux&build=3151)](http://ci.ros2.org/job/ci_linux/3151/)
* Linux-aarch64 [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_linux-aarch64&build=494)](http://ci.ros2.org/job/ci_linux-aarch64/494/)
* macOS [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_osx&build=2508)](http://ci.ros2.org/job/ci_osx/2508/)
* Windows [![Build Status](http://ci.ros2.org/buildStatus/icon?job=ci_windows&build=3203)](http://ci.ros2.org/job/ci_windows/3203/)

mikaelarguedas:
Ok I finally found some time to move this forward.
What this set of PR does:
- move interfaces and corresponding C++/Python fixtures into a new package called test_msgs
- update packages depending on test_communication to use this package
[...]

PR #223, opened 2017-09-01
Test msgs

mikaelarguedas:
connects to #191


test_communication/CMakeLists.txt
```diff
@@ -24,35 +21,27 @@ ament_auto_find_build_dependencies()
...
 if(BUILD_TESTING)
+
+  find_package(ament_cmake REQUIRED)
+  find_package(test_msgs REQUIRED)
+
+  ament_index_get_resource(interface_files "rosidl_interfaces" "test_msgs")
+  string(REPLACE "\n" ";" interface_files "${interface_files}")
+
+  string(REGEX MATCHALL "([A-Z][A-Za-z0-9]*\.msg)" message_files "${interface_files}")
+  string(REGEX MATCHALL "([A-Z][A-Za-z0-9]*\.srv)" service_files "${interface_files}")
```
dirk-thomas: Rather than embedding knowledge about the naming conventions of the interface files I would recommend to iterate over the list and append them to either `message_files` or `service_files` based on the extension  (using [string_ends_with](https://github.com/ament/ament_cmake/blob/d60d7ab800d9046f35b8c68a7d073a712cdae610/ament_cmake_core/cmake/core/string_ends_with.cmake#L25)).

test_communication/CMakeLists.txt
```diff
@@ -24,35 +21,27 @@ ament_auto_find_build_dependencies()
...
-)
-rosidl_generate_interfaces(${PROJECT_NAME}
-  ${message_files}
-  ${other_message_files}
-  ${service_files}
-  DEPENDENCIES builtin_interfaces
-)
-
 if(BUILD_TESTING)
+
```
dirk-thomas: Nitpick: no empty line after `if`.

test_communication/CMakeLists.txt
mikaelarguedas: Yeah I liked this approach because I could get all matches in one invocation without having to iterate. (I started with list(FILTER REGEX) until I realized that it's only available since CMake 3.6...)

I'll change that to iterate on the list manually then

test_communication/CMakeLists.txt
mikaelarguedas: fixed in https://github.com/ros2/system_tests/pull/223/commits/db21f8b81f1a3255c09c74ab7a7f57d5e255d310

test_communication/CMakeLists.txt
mikaelarguedas: addressed in https://github.com/ros2/system_tests/pull/223/commits/66a7db66f03e1c13962ae806a9da4fcfe9b175da

PR #224, opened 2017-09-02
apply forward slash for list_parameters

Karsten1987:
connects to ros2/rclcpp#367


mikaelarguedas:
@Karsten1987 Tip: For `connects to` to work in waffle you need to use the github notation `ros2/rclcpp#367` instead of the full URL

PR #226, opened 2017-09-05
Update test_messages_c.cpp

michielb-prismtech:
Array initialized with 2 while 3 elements filled, increased size.


mikaelarguedas:
@michielb-prismtech Thanks for catching this

PR #230, opened 2017-09-28
Test connext secure

mikaelarguedas:
This enable security tests for Connext.
Not ready to be merged yet given that it seems to work only with the full version of Connext and not the evaluation version that users get from the RTI website (so currently fails on CI).

The first commit enable the tests for Connext static
The second commit allow the tests to pass when the subscriber times out (not sure why the publisher is not brought down properly causing the test to fail, needs further investigation).
The third commit regenerate all the security fixtures using  ros2/sros2#16

This PR DOESN'T add tests for testing access controle policies in Connext


mikaelarguedas:
still not ready for review yet,

CI to see where we're at on all platforms
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3323/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/585/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2651/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3362/)

mikaelarguedas:
Ok this is finally ready for a first round of review.

Remarks:
- the connext nodes that were not shutting down properly seems to be because connext nodes now need ~3-4 seconds to shut down cleanly and were given only 2 in these tests. They now have 7 seconds to shut down that should hopefully make these tests not flaky in that matter
- The osx job has been ran on mini2 that I modified manually to find the connext libssl libraries properly. I'm waiting for feedback from rti support to know what the proper fix is. In the meantime the workaround to copy the libraries where they are expected to be seems to do the job
- the first commit moves the security tests from `test_communication` to `test_security`, I'll squash all the other commits together after review and before merging and will merge this PR with 2 commits: 1 for relocation 1 for the changes

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3331/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/593/)
* macOS [Build Status](http://ci.ros2.org/view/All/job/test_osx_security/9/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3374/)

test_security/CMakeLists.txt
```diff
@@ -0,0 +1,254 @@
...
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake_auto REQUIRED)
+ament_auto_find_build_dependencies()
+
+option(SECURITY "Activate security" OFF)
```
dirk-thomas: Just to consider for naming this options: a lot of options like this start with `ENABLE_`.

test_security/test/test_secure_publisher.cpp
```diff
@@ -77,8 +77,10 @@ int main(int argc, char ** argv)
   std::shared_ptr<rclcpp::node::Node> node = nullptr;
   try {
     node = rclcpp::Node::make_shared(node_name);
-  } catch (std::runtime_error &) {
-    fprintf(stderr, "should not have thrown!");
+  } catch (std::runtime_error & exc) {
+    std::cerr << "should not have thrown!\n" << exc.what();
```
dirk-thomas: Use `printf` instead.

Same below.

test_security/CMakeLists.txt
mikaelarguedas: This option is the one use by [fastrtps](https://github.com/eProsima/Fast-RTPS/blob/5c910f7643c26f7d4764e4d675f23abeadd3c2ee/CMakeLists.txt#L208) and in our CI so if we change it here we would need to check if fast-rtps has been built with security and pass an additional argument to our ci jobs

mikaelarguedas:
> I don't know if the security tests in test_communication have any history or if it is worthwhile to preserve it. The first commit currently makes that history "unaccessible" when viewing the history of the files in their new location.

Github UI doesnt propagate history of moved files anymore. But the history is still there (e.g. blame [still works](https://github.com/ros2/system_tests/blame/test_connext_secure/test_security/test/test_secure_publisher.cpp) on the entire history)

test_security/test/test_secure_publisher.cpp
mikaelarguedas: In the current test_communication, all c++ executables print to standard output using `std::cout` and error messages using `fprintf`. Which one is preferred ? I'm happy to update accordingly

dirk-thomas: https://github.com/ros2/ros2/wiki/Developer-Guide#c-specific

mikaelarguedas: done in https://github.com/ros2/system_tests/pull/230/commits/a55fe0402dc60dd538c2d44bd6b246be2480f45c I'll open a follow-up PR for the tests in this repo but not part of the new `test_security` package

mikaelarguedas:
I squashed all the follow-up commits to merge this as a 2-commit PR: one for file relocation and one for changes.

Issue #242, opened 2017-10-16
Security "not connecting" tests are flaky


dhood:
http://ci.ros2.org/job/nightly_osx_repeated/863

```
projectroot.test_secure_publisher_subscriber__DynamicArrayNested__rmw_connext_cpp__secure_not_connecting_0
projectroot.test_secure_publisher_subscriber__DynamicArrayNested__rmw_connext_cpp__secure_not_connecting_1
projectroot.test_secure_publisher_subscriber__Empty__rmw_connext_cpp__secure_not_connecting_1
projectroot.test_secure_publisher_subscriber__DynamicArrayNested__rmw_fastrtps_cpp__secure_not_connecting_1
```

They are receiving connections when they're not supposed to. We know that some nodes can take a while to shutdown properly so perhaps it's cross-talk from re-running the tests multiple times. I'll add a namespace to the nodes and see if it fixes it.

mikaelarguedas:
I'm afraid adding namespaces is not an easy fix in this case because while the "not connecting" ones won't connect, we would need to generate different security files for all the ones that "should connect" as their current permissions have the partition "rt_".

Edit: Actually looking at the current permission files they all have wildcards on partitions so namespaces should be ok for now :+1:  let's try it

dhood:
closed by https://github.com/ros2/system_tests/pull/243

PR #243, opened 2017-10-17
Add namespace to avoid crosstalk in security 'no connection' tests

dhood:
Aimed at fixing #242 (but it's not certain that cross-talk was causing the issue)

Standard CI on only testing `test_security` package:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3342/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/601/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2676/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3398/)

This build repeatedly tests them 84 times without failure: http://ci.ros2.org/job/ci_osx/2674/ (at which point the buildfarm machine disconnected). The tests are very infrequently flaky so probably only time will tell if this has actually addressed the issue.

### DIFF
--- a/test_security/test/test_secure_publisher.cpp
+++ b/test_security/test/test_secure_publisher.cpp
@@ -71,12 +71,12 @@ int main(int argc, char ** argv)
   }
   rclcpp::init(argc, argv);
   std::string message = argv[1];
-  std::string name_space = argv[2];
+  std::string namespace_ = argv[2];
   std::string node_name = "publisher";
   std::string topic_name = "chatter";
   std::shared_ptr<rclcpp::node::Node> node = nullptr;
   try {
-    node = rclcpp::Node::make_shared(node_name, name_space);
+    node = rclcpp::Node::make_shared(node_name, namespace_);
   } catch (std::runtime_error & exc) {
     fprintf(stderr, "should not have thrown!\n%s\n", exc.what());
     rclcpp::shutdown();
--- a/test_security/test/test_secure_publisher.cpp
+++ b/test_security/test/test_secure_publisher.cpp
@@ -62,7 +62,7 @@ int8_t attempt_publish(
 
 int main(int argc, char ** argv)
 {
-  if (argc != 2) {
+  if (argc != 3) {
     fprintf(
       stderr,
       "Wrong number of arguments,\n"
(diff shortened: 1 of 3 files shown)


test_security/test/test_secure_publisher.cpp
```diff
@@ -71,11 +71,12 @@ int main(int argc, char ** argv)
   }
   rclcpp::init(argc, argv);
   std::string message = argv[1];
+  std::string name_space = argv[2];
   std::string node_name = "publisher";
   std::string topic_name = "chatter";
   std::shared_ptr<rclcpp::node::Node> node = nullptr;
   try {
-    node = rclcpp::Node::make_shared(node_name);
+    node = rclcpp::Node::make_shared(node_name, name_space);
```
mikaelarguedas: nit: name_space -> node_namespace

mikaelarguedas:
> Use consistent naming: name_space vs. namespace (I would lean to the second one since that is the way the API names the argument).

I think that c/c++ linters would complain if we use language keywords to name variables, hence the suggestion to use `node_namespace` instead

dirk-thomas:
> I think that c/c++ linters would complain if we use language keywords to name variables, hence the suggestion to use `node_namespace` instead

The compiler would complain.

The `rclcpp` API already follows the code style guidelines for this and names the argument [namespace_](https://github.com/ros2/rclcpp/blob/989084b3de84000a15fb4bd5a7d63929001cd769/rclcpp/include/rclcpp/node.hpp#L91)

mikaelarguedas:
> The `rclcpp` API already follows the code style guidelines for this and names the argument `namespace_`

Let's use `namespace_` here then :+1:

mikaelarguedas:
Looking at the console output there still seem to be an issue of nodes not being shut down properly.
For example on the MacOS job (test 19 elapsed time 00:48:43) there are 4 dangling nodes (that get denied access to the topic so no crosstalk) on top of the one being tested

dhood:
> Looking at the console output there still seem to be an issue of nodes not being shut down properly.

Pretty sure that we're on the same page, but to be explicit for time-travelers: this PR is a workaround for the wider issue of nodes not shutting down, so it's expected that old nodes are still around since it didn't attempt to address that.

From your investigations it sounds like the context in which tests used to fail now doesn't impact them, so if crosstalk was the only issue then we shouldn't see these tests fail again. If we do, well, there's something else to investigate.

standard CI linux [Build Status](http://ci.ros2.org/job/ci_linux/3347/)

PR #244, opened 2017-10-17
Wait for service before calling it

sloretz:
Wait for the service to be available before calling it.

### DIFF
--- a/test_communication/test/requester_py.py
+++ b/test_communication/test/requester_py.py
@@ -16,41 +16,44 @@ import argparse
 import importlib
 import sys
 
-import rclpy
-from rclpy.node import Node
-from test_msgs.service_fixtures import get_test_srv
 
+def requester(service_name):
+    import rclpy
+    from test_msgs.service_fixtures import get_test_srv
+    rclpy.init(args=[])
 
-class Requester(Node):
-
-    def __init__(self, service_name):
-        super().__init__('requester')
+    try:
         # Import the service
         service_pkg = 'test_msgs'
         module = importlib.import_module(service_pkg + '.srv')
-        self.srv_mod = getattr(module, service_name)
+        srv_mod = getattr(module, service_name)
 
-        self.srv_fixtures = get_test_srv(service_name)
-        self.service_name = 'test_service_' + service_name
+        srv_fixtures = get_test_srv(service_name)
+        service_name = 'test_service_' + service_name
 
-    def run(self):
-        # wait for the service to be available
-        client = node.create_client(self.srv_mod, self.service_name)
-        tries = 15
-        while rclpy.ok() and not client.wait_for_service(timeout_sec=1.0) and tries > 0:
-            print('service not available, waiting again...')
-            tries -= 1
-        assert tries > 0
+        node = rclpy.create_node('requester')
+        try:
+            # wait for the service to be available
+            client = node.create_client(srv_mod, service_name)
+            tries = 15
+            while rclpy.ok() and not client.wait_for_service(timeout_sec=1.0) and tries > 0:
+                print('service not available, waiting again...')
+                tries -= 1
+            assert tries > 0
 
-        print('requester: beginning request')
-        # Make one call to that service
-        for req, resp in self.srv_fixtures:
-            client.call(req)
-            client.wait_for_future()
-            assert repr(client.response) == repr(resp), \
-                'unexpected response %r\n\nwas expecting %r' % (client.response, resp)
-            print('received reply #%d of %d' % (
-                self.srv_fixtures.index([req, resp]) + 1, len(self.srv_fixtures)))
+            print('requester: beginning request')
+            # Make one call to that service
+            for req, resp in srv_fixtures:
+                client.call(req)
+                client.wait_for_future()
+                assert repr(client.response) == repr(resp), \
+                    'unexpected response %r\n\nwas expecting %r' % (client.response, resp)
+                print('received reply #%d of %d' % (
+                    srv_fixtures.index([req, resp]) + 1, len(srv_fixtures)))
+        finally:
+            node.destroy_node()
+    finally:
+        rclpy.shutdown()
 
 
 if __name__ == '__main__':
@@ -59,16 +62,9 @@ if __name__ == '__main__':
                         help='name of the ROS message')
     args = parser.parse_args()
     try:
-        rclpy.init(args=[])
-        node = Requester(args.service_name)
-        try:
-            node.run()
-        finally:
-            node.destroy_node()
+        requester(service_name=args.service_name)
     except KeyboardInterrupt:
         print('requester stopped cleanly')
     except BaseException:
         print('exception in requester:', file=sys.stderr)
         raise
-    finally:
-        rclpy.shutdown()
--- a/test_communication/test/requester_py.py
+++ b/test_communication/test/requester_py.py
@@ -20,17 +20,17 @@ import sys
 def requester(service_name):
     import rclpy
     from test_msgs.service_fixtures import get_test_srv
+
+    # Import the service
+    service_pkg = 'test_msgs'
+    module = importlib.import_module(service_pkg + '.srv')
+    srv_mod = getattr(module, service_name)
+
+    srv_fixtures = get_test_srv(service_name)
+    service_name = 'test_service_' + service_name
+
     rclpy.init(args=[])
-
     try:
-        # Import the service
-        service_pkg = 'test_msgs'
-        module = importlib.import_module(service_pkg + '.srv')
-        srv_mod = getattr(module, service_name)
-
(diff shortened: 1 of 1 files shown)


sloretz:
@mikaelarguedas It looks like the console output is no longer an issue on linux. What kind of error messages only appear if the node shuts down before the test timeout?

I tested console output by adding
```python
    while rclpy.ok():
        print('still sleeping...')
        time.sleep(1)
```

Which appeared to output the message `still sleeping...` in real time.
```
25: (test_requester) pid 24547: ['/usr/bin/python3', '/workspace/ros2_ws/src/ros2/system_tests/test_communication/test/requester_py.py', 
'Empty'] (stderr > stdout, all > console)
25: [test_requester] still sleeping...
25: [test_replier] replier: beginning loop
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 2
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 3
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 4
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 5
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 6
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 7
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 8
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 9
25: [test_requester] still sleeping...
25: [test_requester] still sleeping...
25: [test_replier] spin_count: 10
25: [test_requester] still sleeping...
25: (test_replier) rc 0
25: () tear down
25: (test_requester) signal SIGINT
25: [test_requester] requester stopped cleanly
25: (test_requester) rc 0
25: F
```

mikaelarguedas:
> I think this is not an issue anymore but am not sure.
We used to prevent our nodes tests from timing out using a condition like the number_of_cycles in the process with the main exit handler to make sure our nodes have time to shutdown properly before being terminated (and also get console output and more explicit error messages). The console output concern should be solved now that we use PYTHONUNBUFFERED for all these tests. Not sure about the other points though

@sloretz FYI: it appears that the printing issue reappeared now that we moved to pytest (https://github.com/ament/ament_cmake/pull/118). I also believe that the issue of node not shutting down properly when being killed is still present (https://github.com/ros2/system_tests/pull/243).

sloretz:
@mikaelarguedas 584b792616a7ae2fb72c51037702a581db378f9b waits a maximum of 15 seconds so the client can shutdown cleanly

test_communication/test/requester_py.py
```diff
@@ -32,6 +32,11 @@ def requester(service_name, number_of_cycles):
     srv_fixtures = get_test_srv(service_name)
 
     client = node.create_client(srv_mod, 'test_service_' + service_name)
+    tries = 15
+    while rclpy.ok() and not client.wait_for_service(timeout_sec=1.0) and tries > 0:
+        print('service not available, waiting again...')
+        tries -= 1
+    assert tries > 0
```
mikaelarguedas: We actually need to cleanup/shutdown here before exiting.
```
     node.destroy_node()
     rclpy.shutdown()
```

sloretz: @mikaelarguedas refactored to make sure cleanup happens in cb8ce48cff29dec596249db096b9dc29b16af6b3

I removed `--number_of_cycles`/`spin_count` since they were broken and unused. The while loop only executed once because of a `break` statement. `grep` in `system_tests` for `number_of_cycles` and didn't find anything using it.

sloretz:
CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3645/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/825/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2979/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3746/)

test_communication/test/requester_py.py
mikaelarguedas: @sloretz indeed number of cycles was not used here :+1: I guess it was a remnant of the original copy-n-paste from the publisher subscriber.

mikaelarguedas:
Not sure that refactoring the tests to inherit from node belong in this PR / should be done only on that file.
Same comment regarding the use of logging. These tests being meant to test communication, they should restrict themselves to do only this and not introduce other components

sloretz:
@mikaelarguedas I removed the logging calls in 61f5c5a. It sounds like you're saying if the tests should inherit from node then it should be done in a separate PR where all files are updated to keep the code consistent.  How about I switch it back to a `requester` function, and move the `try/finally` cleanup code into it?

mikaelarguedas:
> How about I switch it back to a requester function, and move the try/finally cleanup code into it?

Sounds good to me

sloretz:
@mikaelarguedas Back to plain function in c6ca8d604e72301d0a415cedf6b922092a77fbc1

CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3649/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/829/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2986/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3751/)

test_communication/test/requester_py.py
```diff
@@ -17,55 +17,52 @@
...
-    node = rclpy.create_node('requester')
+    try:
+        # Import the service
+        service_pkg = 'test_msgs'
+        module = importlib.import_module(service_pkg + '.srv')
+        srv_mod = getattr(module, service_name)
 
-    srv_fixtures = get_test_srv(service_name)
+        srv_fixtures = get_test_srv(service_name)
+        service_name = 'test_service_' + service_name
```
mikaelarguedas: Everything before `create_node` can stay above `rclpy.init()` and thus be outside of the try block (except the get_test_srv call maybe)

test_communication/test/requester_py.py
```diff
@@ -17,55 +17,52 @@
...
-    client = node.create_client(srv_mod, 'test_service_' + service_name)
+        node = rclpy.create_node('requester')
+        try:
+            # wait for the service to be available
+            client = node.create_client(srv_mod, service_name)
+            tries = 15
+            while rclpy.ok() and not client.wait_for_service(timeout_sec=1.0) and tries > 0:
+                print('service not available, waiting again...')
+                tries -= 1
+            assert tries > 0
```
mikaelarguedas: can you please add an assert message here ? so that it's easy to now what went wrong if we have only stderr output

test_communication/test/requester_py.py
```diff
@@ -17,55 +17,52 @@
...
-    node.destroy_node()
-    rclpy.shutdown()
-    assert len(received_replies) == len(srv_fixtures), \
-        'Should have received %d responses from replier' % len(srv_fixtures)
-    print('everything went well !')
+            print('requester: beginning request')
+            # Make one call to that service
+            for req, resp in srv_fixtures:
+                client.call(req)
+                client.wait_for_future()
```
mikaelarguedas: if this doesn't return in a timely manner we still have potential for timing out without cleanup. That was the problem before this PR and is out of scope but should be ticketed so that we address it soon

test_communication/test/requester_py.py
sloretz: Looks like `get_test_srv` works fine before `rclpy.init()`
d5afa0e46ba37a74bf84e9b5c15fc5d6f6eca1db

test_communication/test/requester_py.py
sloretz: Done! d5afa0e46ba37a74bf84e9b5c15fc5d6f6eca1db

sloretz:
New CI
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3650/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/830/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2987/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3752/)

mikaelarguedas:
@sloretz I think that the reason for the MacOS test failures is that this branch has not been rebased so it doesnt have the fix from https://github.com/ros2/system_tests/pull/250

sloretz:
@mikaelarguedas Rebased and re-running OSX CI http://ci.ros2.org/job/ci_osx/2995/

mikaelarguedas:
@sloretz CI is happy, my approval still stands feel free to merge

PR #245, opened 2017-11-14
use ament_cmake_pytest instead of ament_cmake_nose

dirk-thomas:
Connect to ament/ament_cmake#115.

### DIFF
--- a/test_communication/package.xml
+++ b/test_communication/package.xml
@@ -20,8 +20,8 @@
   <exec_depend>rosidl_default_runtime</exec_depend>
 
   <test_depend>ament_cmake</test_depend>
-  <test_depend>ament_cmake_nose</test_depend>
   <test_depend>ament_cmake_gtest</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
   <test_depend>launch</test_depend>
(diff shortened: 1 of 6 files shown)


test_communication/package.xml
```diff
@@ -20,7 +20,7 @@
   <exec_depend>rosidl_default_runtime</exec_depend>
 
   <test_depend>ament_cmake</test_depend>
-  <test_depend>ament_cmake_nose</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
```
mikaelarguedas: Nit: alphabetical order

test_security/package.xml
```diff
@@ -12,7 +12,7 @@
   <buildtool_depend>ament_cmake_auto</buildtool_depend>
 
   <test_depend>ament_cmake</test_depend>
-  <test_depend>ament_cmake_nose</test_depend>
+  <test_depend>ament_cmake_pytest</test_depend>
```
mikaelarguedas: Nit: alphabetical order

test_communication/package.xml
dirk-thomas: I didn't notice that they were out-of-order before already. Anyway fixed.

test_security/package.xml
dirk-thomas: I didn't notice that they were out-of-order before already. Anyway fixed.

PR #247, opened 2017-11-27
Extend service test suite

mikaelarguedas:
connects to ros2/rcl_interfaces#25

should be merged after #244

### DIFF
--- a/test_communication/test/test_replier.cpp
+++ b/test_communication/test/test_replier.cpp
@@ -74,6 +74,14 @@ int main(int argc, char ** argv)
 
   auto services_empty = get_services_empty();
   auto services_primitives = get_services_primitives();
+  auto services_dynamic_array_primitives = get_services_dynamic_array_primitives();
+  auto services_dynamic_array_primitives_nested = get_services_dynamic_array_primitives_nested();
+  auto services_bounded_array_primitives = get_services_bounded_array_primitives();
+  auto services_static_array_primitives = get_services_static_array_primitives();
+  auto services_nested = get_services_nested();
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
ros2/system_tests#244 has merged, placing this in review

Note that this will considerably increase testing time so we may want to discuss the testing strategy before merging. But I'd like reviews in the meantime

mikaelarguedas:
>  Can you post some numbers for comparison.

I'll rerun a set of jobs once #248 and #249 are merged and now that I can use wait for service rather than a hard coded 2-second sleep to have more reliable numbers.

A regular run of test_communication (running once each test without repeats) is currently 241 tests for ~15min.
On the job I used for testing this [Build Status](http://ci.ros2.org/job/ci_linux/3614/) it took ~20min and I expect it to grow to >30min the day we start doing cross-vendor service testing.

Note: these number assume we use only 2 rmw for testing. This will also be much longer if we readd openspllice to our regular CI

mikaelarguedas:
More numbers (based on the jobs posted in the previous comment):
Before this PR: ~15min to run 241 tests from test_communication (5h06min for 21 runs during the repeated nighties)
After this PR: ~23min to run 369 tests (~2h20min for 6 runs). Estimated repeated nightlies time: ~8h30min

nuclearsandwich:
@hidmic I know you're looking at increasing system test coverage. This PR seems like something we should revisit and either get merged or decide to retire.

hidmic:
Closing for the same [reasons](https://github.com/ros2/rcl_interfaces/pull/25#issuecomment-514329373) https://github.com/ros2/rcl_interfaces/pull/25 was closed.

PR #250, opened 2017-11-29
use brew to find openssl library path and add it to the library dirs

mikaelarguedas:
Move logic adding openssl libraries to `DYLD_LIBRARY_PATH` from the ci script into this package (the only one using connext security)

connects to ros2/ci#112


mikaelarguedas:
> I think it deserves a comment of why we do this though

Sounds good I'll add that before placing it in review :+1:

mikaelarguedas:
(note to self) consider factoring out this logic if we happen to need it in other places

mikaelarguedas:
+comment I forgot to add : https://github.com/ros2/system_tests/commit/684b442d528fe895fd1fc98c083e83f76cc23b95

PR #253, opened 2017-12-03
Regression test for set_parameters with bad callback reference

dhood:
Connects to ros2/rclcpp#414
Regression test for https://github.com/ros2/rclcpp/pull/414

Without https://github.com/ros2/rclcpp/pull/414, the test will crash (I'm getting `C++ exception with description "std::bad_alloc" thrown in the test body.` but in https://github.com/ros2/demos/pull/203/files#diff-a5a151889abb3214250ba9db05358e4bR92 it would segfault).

CI including ros2/rclcpp#414
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3695/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/862/)
* macOS skipped because queue is busy
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3799/)

### DIFF
--- a/test_rclcpp/test/test_local_parameters.cpp
+++ b/test_rclcpp/test/test_local_parameters.cpp
@@ -103,6 +103,41 @@ TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), local_asynchronous) {
   verify_get_parameters_async(node, parameters_client);
 }
 
+TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), local_async_with_callback) {
+  using rclcpp::parameter::ParameterVariant;
+  using SetParametersResult =
+      std::shared_future<std::vector<rcl_interfaces::msg::SetParametersResult>>;
+
+  auto node = rclcpp::Node::make_shared("test_parameters_local_async_with_callback");
+  // TODO(esteve): Make the parameter service automatically start with the node.
+  auto parameter_service = std::make_shared<rclcpp::parameter_service::ParameterService>(node);
+  auto parameters_client = std::make_shared<rclcpp::parameter_client::AsyncParametersClient>(node);
+  if (!parameters_client->wait_for_service(20s)) {
+    ASSERT_TRUE(false) << "service not available after waiting";
+  }
+  rclcpp::executors::SingleThreadedExecutor executor;
+  auto set_parameters_results = parameters_client->set_parameters({
+    ParameterVariant("foo", 2),
+    ParameterVariant("bar", "hello"),
+    ParameterVariant("barstr", std::string("hello_str")),
+    ParameterVariant("baz", 1.45),
+    ParameterVariant("foobar", true),
+    ParameterVariant("barfoo", std::vector<uint8_t>{3, 4, 5}),
+  },
+    [&executor](SetParametersResult future)
+    {
+        printf("Got set_parameters result\n");
+        // Check to see if they were set.
+        for (auto & result : future.get()) {
+          ASSERT_TRUE(result.successful);
+        }
+        executor.cancel();
+    }
+  );
+  executor.add_node(node);
+  executor.spin();
+}
+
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_local_parameters.cpp
```diff
@@ -103,6 +103,61 @@ TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), local_asynchronous) {
...
+// go out of scope before it gets called: see https://github.com/ros2/rclcpp/pull/414
+TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), local_async_with_callback) {
+  auto node = std::make_shared<ParametersAsyncNode>();
+  // TODO(esteve): Make the parameter service automatically start with the node.
+  auto parameter_service = std::make_shared<rclcpp::parameter_service::ParameterService>(node);
+  // TODO(dhood): Make the parameter client constructable from Node subclass' constructor.
+  node->parameters_client_ =
+    std::make_shared<rclcpp::parameter_client::AsyncParametersClient>(node);
+  if (!node->parameters_client_->wait_for_service(20s)) {
+    ASSERT_TRUE(false) << "service not available after waiting";
```
mikaelarguedas: I saw that you copied it from above but do you know why we don't just use `FAIL()` here?
`FAIL() << "service not available after waiting";`

dhood: perhaps it doesn't allow a message (it's only used as `FAIL()` elsewhere FWICS)

mikaelarguedas: Yeah we can look into what the differences are at a future date.

Note to future-self @mikaelarguedas: consider replacing those in the future by `FAIL()` or `ADD_FAILURE()` based on context.
references: [places where we use it](https://github.com/ros/class_loader/blob/bf5ed3ab137eef08dede09ecc039ac37a40b440c/test/utest.cpp#L330), [gtest example](https://github.com/google/googletest/blob/master/googletest/docs/AdvancedGuide.md#explicit-success-and-failure)

dhood:
this PR still needs review

CI OS X including https://github.com/ros2/system_tests/pull/253/commits/a1ea944e159a2385f211901c5a01d6b2c96c0085: [Build Status](http://ci.ros2.org/job/ci_osx/3039/)

test_rclcpp/test/test_local_parameters.cpp
```diff
@@ -103,6 +103,60 @@ TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), local_asynchronous) {
...
+  // TODO(esteve): Make the parameter service automatically start with the node.
+  auto parameter_service = std::make_shared<rclcpp::parameter_service::ParameterService>(node);
+  if (!node->parameters_client_->wait_for_service(20s)) {
+    ASSERT_TRUE(false) << "service not available after waiting";
+  }
+  rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(node);
+  node->queue_set_parameter_request(executor);
+  // When the parameters client receives its response it will cancel the executor.
+  executor.spin();
```
wjwwood: The only issue here is that there is no timeout for the spin. If the callback is not called the test will hang.

You could instead:

- create a promise and a future from the promise
- pass the promise to `queue_set_parameter_request`
- capture the promise in the lambda callback
- set the promise in the lambda callback
- `spin_until_future_complete` on the future and with a timeout
- assert the future is set after `spin_until_future_complete` returns (i.e. assert it didn't timeout)

dhood: Thanks for the idea, I wasn't aware of this way to make the test fail in a way other than timeout: it's a better approach since this test file has a really long time out because of all the `wait_for_service` calls. I won't have time to update this tonight (and the test shouldn't be failing as it stands) so I'll still merge as is for now

PR #263, opened 2018-04-18
Place RTI OpenSSL on the (LD_LIBRARY_)PATH on Linux

mikaelarguedas:
This implements "option a" described in https://github.com/ros2/ros2/issues/481#issuecomment-381796417 as option b was evaluated as not acceptable for users.

This also works on xenial as the env vars stay the same so it can be merged async from https://github.com/ros2/ci/pull/148 (but there's not much incentive to merge it before)

Xenial job: [Build Status](http://ci.ros2.org/job/ci_linux/4258/)
Bionic job: [Build Status](http://ci.ros2.org/job/ci_linux/4256/)

connects to ros2/ros2#481


mikaelarguedas:
This is broken on windows, moving back to "in progress" for now

mikaelarguedas:
I changed this to modify the PATH only when testing Connext on Linux to make sure we still exercise OpenSSL 1.1 when using Fast-RTPS.
This doesn't modify the PATH for Windows and MacOS but we could use RTI's OpenSSL libraries on these platforms by just removing [this condition](https://github.com/ros2/system_tests/pull/263/files#diff-6fc729fbfd434d3b129f70533aa0346dR262). As there is no need for now I didnt do it.
If we were to test RTI's OpenSSL on other platforms we would also need to modify [this logic](https://github.com/ros2/system_tests/pull/263/files#diff-6fc729fbfd434d3b129f70533aa0346dR292) to modify the *LIBRARY_PATH on all platforms to include RTI's OpenSSL libraries.

CI with last changes:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4356/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1329/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3567/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4420/)

* Bionic [Build Status](http://ci.ros2.org/job/ci_linux/4355/)

mikaelarguedas:
> Reading the patch I was confused about TEST_PATH_WITH_RTI_BIN assuming it would contain the extra bin for all rmw. Further down I see that it is only populated in the case of Connext. Maybe renaming the variable to something like TEST_PATH would avoid that confusion.

Good point I didnt rename the variable after changing the logic: done in 6dbd6102f72dc44ad27109f6260737f48fa15e16

PR #266, opened 2018-05-01
Regression test for FastCDR typesupport.

mjcarroll:
Adds a regression test for the the bug we saw in Bionic with FastRTPS.

Connects to: ros2/rclcpp#464


mjcarroll:
@mikaelarguedas I don't know if this is the right place for this to live.  It's really only testing FastCDR, but it also depends on rcl_interfaces and rclcpp.

mjcarroll:
I would be good with dropping the test entirely, this is essentially what I needed to completely isolate the issue, but probably doesn't have enough coverage to justify standing on it's own.

mjcarroll:
Also, with another look, it doesn't look like it actually depends on rclcpp, that was leftover from my exploration.

PR #269, opened 2018-05-08
account for null-terminator character

mikaelarguedas:
This fixes a [gcc warning](https://ci.ros2.org/job/ci_packaging_linux/92/warnings23Result/package.-1552620408/) as the string to store can be up to 4 numeric characters + the null-terminator so the container should be of size 5.

related to https://github.com/ros2/ros2/issues/481


mikaelarguedas:
Because we know the [value of `size`](https://github.com/ros2/system_tests/blob/600d5cac98d947f40500ea4e2f0092c45bf3fec4/test_communication/test/test_messages_c.cpp#L583). I agree that it's fragile though.

I'm not aware of a macro returning the maximum length of a string representation of a type, (but if there's one I'm more than happy to use it!).
Otherwise, I can assume that for the foreseeable future we will have systems that have at most 64bits to represent size_t and use 21 for the array: 
```
In [0]: len(str(2**64))
Out[0]: 20
# + null-terminator
```
Does that match your suggestion?

dirk-thomas:
That sounds better to me :+1:

PR #272, opened 2018-05-31
Pass initial parameter values to node constructor

sloretz:
~~Needs to be targeted at master after #271 is merged.~~ Targeted at master

connects to ros2/rclcpp#486


mikaelarguedas:
as this and connected PRs were approved before the [renaming of "initial_values" to "initial_parameters"](https://github.com/ros2/rclcpp/pull/486/commits/29046af5d29a085c4edb81a698bdac9124978823). Maybe we should apply the same variable renaming to all the connected PRs

PR #274, opened 2018-06-05
Test initializing parameters from command line

sloretz:
Requires #272 
Connects to ros2/rclcpp#488

This adds tests for parameters initialized via the command line arg `__params:=/path/to/yaml.file`


wjwwood:
https://github.com/ros2/system_tests/pull/274/commits/41a5dcca6be4928e77cca50b7081abfbb56f899d 👍

sloretz:
@mikaelarguedas merged your branch into this one; thanks for the fixes!

sloretz:
Limited CI to test if windows issue is fixed: [Build Status](https://ci.ros2.org/job/ci_windows/4643/)

sloretz:
Another rebuild to check if windows issue is fixed [Build Status](https://ci.ros2.org/job/ci_windows/4644/)

sloretz:
One more windows CI attempt [Build Status](https://ci.ros2.org/job/ci_windows/4645/)

sloretz:
Yet another windows CI attempt

[Build Status](https://ci.ros2.org/job/ci_windows/4650/)

sloretz:
All platforms CI testing this package

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4566/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1476/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3753/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4652/)

sloretz:
@mikaelarguedas here's a summary of the changes to this PR

1.  Added custom `NamedTemporaryFile` class to util to work around issue with `tempfile.NamedTemporaryFile()` on windows 
    * A temporary file on windows can't be opened by another process without closing it, but it can't be closed without being deleted. The workaround is to create a temporary directory with a normal file inside, then delete the whole directory on `__exit__()`.
1. Got rid of `asycnio` and coroutines.
    * I was having trouble debugging the next issue, so I rewrote the tests using `subprocess` and `threading`
1. Every helper executable starts with a different node name
    * I'm not sure what the real bug is, but I don't have time to get to root cause right now. Here's what I know.
        * Every test does the following:
            1. spawns an executable with service server `/initial_params_node/get_parameters`
            1. creates a client for that service
            1. calls `wait_for_service()`
            1. calls the service and checks values of parameters returned
            1. destroys the client
            1. shuts down the executable
        * The first test to run passes, but the second test never gets a response from the service server.
        * The issue seems to be lower level than `rclpy`. The second test's client handle is in the wait set but never becomes ready. I did not check with wireshark to see if the request was actually sent.
        * It's definitely a race condition, adding `time.sleep(1)` after calling `wait_for_service()` but before calling the service in the second test solves the issue.
        * Reusing the same client for all tests does not solve the issue.
        * Giving each test a different node name (and therefor a different service name) works around the issue

test_cli/test/utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def __enter__(self):
+        self._proc = subprocess.Popen(self._command, env=self._env)
+        return self
+
+    def __exit__(self, t, v, tb):
+        self._proc.kill()
+
+
+class NamedTemporaryFile:
```
dirk-thomas: The naming of this custom class is kind of confusing. It is the same as the one from the Python library but it behaves differently (doesn't return a file handle but the name). Therefore I would suggest to rename it to something more "custom" to avoid confusion.

dirk-thomas:
> I'm not sure what the real bug is, but I don't have time to get to root cause right now.

Please ticket this problem to investigate it in the future.

test_cli/test/utils.py
sloretz: `TemporaryFileWithContent` in deb59989c82059923b1fcfa9436476c7e12b9aec

sloretz:
> Please ticket this problem to investigate it in the future.

ros2/ros2#502

test_cli/package.xml
```diff
@@ -0,0 +1,27 @@
...
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake_auto</buildtool_depend>
+
+  <build_depend>ament_cmake</build_depend>
+
+  <test_depend>ament_cmake_pytest</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>launch</test_depend>
```
wjwwood: Seems like an unused dependency?

sloretz:
@wjwwood The intent was to combine `test_cli_remapping` with testing parameters into `test_cli`, but #268 was merged first. I'll leave launch as a dependency for the moment and add it as a note to a ticket to combine the two test packages.

dirk-thomas:
The tests introduced in this PR are failing on the Windows debug nightly: https://ci.ros2.org/view/nightly/job/nightly_win_deb/906/

PR #284, opened 2018-06-17
only test serialization on fastrtps and connext

Karsten1987:
Run only the serialized functions on fastrtps and connext. This should make the opensplice builds happy.

CI only on Linux with Fastrtps and Opensplice. Testing only `test_comunication`.
[Build Status](https://ci.ros2.org/job/ci_linux/4706/)


dirk-thomas:
How about `rmw_connext_dynamic_cpp`?

mikaelarguedas:
> How about `rmw_connext_dynamic_cpp`?

I don't think any of the CDR things have been tested for connext dynamic so my guess is that this would not work.
As we never build connext dynamic I don't think it even compiles at the moment

test_communication/CMakeLists.txt
```diff
@@ -315,13 +315,19 @@ if(BUILD_TESTING)
   endif()
 
   macro(serialize)
+    if(${rmw_implementation} STREQUAL "rmw_connext_dynamic_cpp" OR ${rmw_implementation} STREQUAL "rmw_opensplice_cpp")
+      message(STATUS "skipping serialize tests for ${rmw_implementation}")
+      set(SKIP_TEST "SKIP_TEST")
```
mikaelarguedas: this should be set to an empty string by default, otherwise the variable will keep the same value and the following tests will be skipped as well (same in the other macro below)

Karsten1987: done in f6ca0ed

Karsten1987:
> How about rmw_connext_dynamic_cpp?
I personally feel strong about removing `rmw_connext_dynamic_cpp` all together. However, if wished I am happy to fill out the same empty messages as for opensplice.

dirk-thomas:
> if wished I am happy to fill out the same empty messages as for opensplice.

Please go ahead and do that since it is a minimal task and avoids further divergence.

mikaelarguedas:
The PR now skips connext_dynamic explicitly: https://github.com/ros2/system_tests/pull/284/commits/610b0e5de5f64fbf9b760f59790dd998d6702ac8

dirk-thomas:
I was referring to a similar change as ros2/rmw_opensplice#228 for `rmw_connext_dynamic_cpp`.

mikaelarguedas:
> I was referring to a similar change as ros2/rmw_opensplice#228 for rmw_connext_dynamic_cpp.

oh my bad I misunderstood :+1:

PR #287, opened 2018-07-13
disable xmllint for now

dirk-thomas:
Skips the linter introduced in ament/ament_lint#104 for now.

### DIFF
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -44,6 +44,11 @@ if(BUILD_TESTING)
     # endif()
   endforeach()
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
The xml files were still linted in the last nightlies and therefore failed the test, e.g. https://ci.ros2.org/view/nightly/job/nightly_linux_debug/lastCompletedBuild/testReport/

It seems to work for me locally. Looking into it...

dirk-thomas:
Commit e2875de859489e75501e4994369a92510e3d26d1 disables xmllint correctly.

PR #293, opened 2018-08-29
Test parameter behavior for rclpy nodes

nuclearsandwich:
Adds an rclpy node to the test fixtures so that our tests cover both client libraries.

For these tests to succeed they must be run along with https://github.com/ros2/rclpy/pull/225

Connects to ros2/rclpy#202

### DIFF
--- a/test_cli/CMakeLists.txt
+++ b/test_cli/CMakeLists.txt
@@ -37,7 +37,7 @@ if(BUILD_TESTING)
     PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
     ENV
       INITIAL_PARAMS_RCLCPP=$<TARGET_FILE:initial_params_rclcpp>
-      INITIAL_PARAMS_RCLPY=test/initial_params.py
+      INITIAL_PARAMS_RCLPY=${CMAKE_CURRENT_LIST_DIR}/test/initial_params.py
   )
   set_tests_properties(test_params_yaml
     PROPERTIES DEPENDS
--- a/test_cli/CMakeLists.txt
+++ b/test_cli/CMakeLists.txt
@@ -37,6 +37,7 @@ if(BUILD_TESTING)
     PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
     ENV
       INITIAL_PARAMS_RCLCPP=$<TARGET_FILE:initial_params_rclcpp>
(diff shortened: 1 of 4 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5114/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1894/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4240/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5091/)

First round of CI. There's a supplemental repos file that includes the necessary rclpy branch.

test_cli/test/utils.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, command):
 
         # Execute python files using same python used to start this test
         if command[0][-3:] == '.py':
+            self._command = list(self._command)
```
nuclearsandwich: This was needed because the tuples that get passed as commands are immuntable.

test_cli/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ if(BUILD_TESTING)
     PYTHON_EXECUTABLE "${_PYTHON_EXECUTABLE}"
     ENV
       INITIAL_PARAMS_RCLCPP=$<TARGET_FILE:initial_params_rclcpp>
+      INITIAL_PARAMS_RCLPY=test/initial_params.py
```
sloretz: I'm not sure what I would have expected the working directory to be when the test is run. If this passes I guess that means it is the top level source directory containing `CMakeLists.txt`.

nuclearsandwich: But I think it's also reasonable to set the source directory either via `CMAKE_CURRENT_LIST_DIR` or some other preferred variable for source paths.

nuclearsandwich:
Testing again using CMAKE_CURRENT_LIST_DIR to get the absolute source directory path rather than leaving it to the chance of the current directory.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5118/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1898/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4244/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5096/)

PR #296, opened 2018-09-06
Add in a test for the new get_parameter_or_set_default API.

clalancette:
This should only be merged if https://github.com/ros2/rclcpp/pull/551 is approved and merged.


test_rclcpp/test/test_local_parameters.cpp
```diff
@@ -414,6 +414,41 @@ TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), get_parameter_or) {
...
+
+  // Check to see if they were set.
+  for (auto & result : set_parameters_results) {
+    ASSERT_TRUE(result.successful);
+  }
+
+  {
+    // try to get with default a parameter that is already set
+    int64_t foo_value = -1;
+    node->get_parameter_or("foo", foo_value, static_cast<int64_t>(42));
```
sloretz: Is this supposed to be a call to `get_parameter_or_set_default()`?

clalancette: Yes, absolutely, thanks for pointing it out.  Cut-n-paste error.  It should be fixed now.

PR #303, opened 2018-10-29
Handles msg files with the same name in different subfolders

hidmic:
Connected to ros2/rosidl#301. This pull request tweaks `test_communication` and `test_security` packages to properly handle .msg and .srv definition in action folders.

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -28,6 +28,11 @@ if(BUILD_TESTING)
   set(message_files "")
   set(service_files "")
   foreach(interface_file ${interface_files})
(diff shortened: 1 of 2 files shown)


hidmic:
Nothing to add, you're spot on. Only downside is that we lose test coverage for generated action messages and services until we get to the point where we can write tests for actions. Changes applied. PTAL.

apojomovsky:
Missed this one as part of yesterday's merge group. Merging now.

dirk-thomas:
@hidmic @apojomovsky Is the branch still needed or can it be deleted?

apojomovsky:
It can be safely deleted @dirk-thomas . My mistake here, I should have deleted it after merging.

PR #304, opened 2018-10-29
Get parameter map

clalancette:
This PR adds tests for the get/set parameter map functionality in https://github.com/ros2/rclcpp/pull/575.  See that PR for more description on the functionality, CI, and links to issues.

### DIFF
--- a/test_rclcpp/test/test_local_parameters.cpp
+++ b/test_rclcpp/test/test_local_parameters.cpp
@@ -477,6 +477,40 @@ TEST(CLASSNAME(test_local_parameters, RMW_IMPLEMENTATION), set_parameter_if_not_
     ASSERT_TRUE(node->get_parameter("bar", bar_value));
     ASSERT_EQ(bar_value, 42);
   }
+
+  {
+    // try to set a map of parameters
+    std::map<std::string, double> bar_map{
+      {"x", 0.5},
+      {"y", 1.0},
+    };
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Is there a reason why these tests can't be directly in `rclcpp`?

clalancette:
> Is there a reason why these tests can't be directly in `rclcpp`?

Not particularly; it does strike me as odd that they are in `system_tests`, but that's where the other `get_parameter` tests and friends are: https://github.com/ros2/system_tests/blob/master/test_rclcpp/test/test_local_parameters.cpp#L381 (and below).  I could move the tests for this PR into rclcpp/test if we think that is better, then do another PR to move the other tests here as well.

wjwwood:
Historically, I think `test_rclcpp` exists because we could not select the rmw implmentation within rclcpp, so if you wanted to run the tests across all rmw implementations you needed to do it in a separate package.  I don't think that's the case any longer, so ideally everything in `test_rclcpp` could be moved to `rclcpp` again. It's possible that there are still some tests which are better in a separate package, like if it uses the CMake interface provided by `rclcpp`.

dirk-thomas:
I was mostly focus on the rmw implementation independent things - like local node parameters. Those shouldn't run N times (for each rmw) but only once.

clalancette:
OK, closing this one in favor of putting the tests directly into https://github.com/ros2/rclcpp/pull/575

PR #313, opened 2018-11-20
refactor to support init options and context

wjwwood:
Connects to ros2/rmw#154

### DIFF
--- a/test_communication/test/test_messages_c.cpp
+++ b/test_communication/test/test_messages_c.cpp
@@ -21,6 +21,8 @@
 #include <string>
 #include <thread>
 
+#include "osrf_testing_tools_cpp/scope_exit.hpp"
+
 #include "rcl/subscription.h"
 #include "rcl/publisher.h"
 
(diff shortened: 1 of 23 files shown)


test_communication/test/test_messages_c.cpp
```diff
@@ -102,9 +115,11 @@ class CLASSNAME (TestMessagesFixture, RMW_IMPLEMENTATION) : public ::testing::Te
     rcl_ret_t ret = rcl_node_fini(this->node_ptr);
     delete this->node_ptr;
     EXPECT_EQ(RCL_RET_OK, ret) << rcl_get_error_string().str;
-    ret = rcl_shutdown();
+    ret = rcl_shutdown(this->context_ptr);
+    delete this->context_ptr;
```
sloretz: `rcl_context_fini()`

test_rclcpp/CMakeLists.txt
```diff
@@ -66,6 +66,16 @@ if(BUILD_TESTING)
       "rclcpp")
   endfunction()
 
+  function(custom_gtest_executable target)
+    ament_add_gtest_executable(${target} ${ARGN})
+    target_compile_definitions(${target}
+      PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
```
sloretz: `RMW_IMPLEMENTATION` as a compile definition works? I thought it had to be an environment variable. Test communication uses an environment variable

https://github.com/ros2/system_tests/blob/394f326f2b3da52d802dad1090cddd8ef6fba5b7/test_communication/CMakeLists.txt#L94-L102

test_rclcpp/test/test_executor.cpp
```diff
@@ -40,6 +40,7 @@
 using namespace std::chrono_literals;
 
 TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), recursive_spin_call) {
+  if (!rclcpp::ok()) {rclcpp::init(0, nullptr);}
```
sloretz: Why this line instead of `init()` and `shutdown()` in `main()`? It looks like `rclcpp` won't be shut down.

test_rclcpp/CMakeLists.txt
wjwwood: This is a direct copy of `custom_executable` except that it uses `ament_add_gtest_executable` instead of `add_executable`, so I don't know but that seems to be the right behavior.

test_rclcpp/test/test_executor.cpp
wjwwood: Calling shutdown explicitly is not required.

A few of the test files with the init/shutdown in the main actually had one or more tests that called shutdown themselves, causing tests that ran after them to fail because "init had not been called" (since the shutdown in the other test).

I found this to be cleaner, as it ensures init has been called without shutting down and initializing again unnecessarily. Also it is resilient to tests being filtered by gtest.

wjwwood: > Calling shutdown explicitly is not required.

To be clear, you may call shutdown manually, or it is called in the SIGINT signal handler for each Context which elected to have `shutdown_on_sigint=true` (the default) and which are not already shutdown, or finally it is called in the destructor of Context right before finalizing the rcl context object, if not already shutdown.

test_communication/test/test_messages_c.cpp
wjwwood: Good catch, fixed in f06076c!

test_rclcpp/CMakeLists.txt
sloretz: Ah nevermind. It looks like the compile definition is used to set test name

https://github.com/ros2/system_tests/blob/394f326f2b3da52d802dad1090cddd8ef6fba5b7/test_rclcpp/test/test_services_client.cpp#L25-L30

The environment variable is set when `custom_launch_test_two_executables()` is called,

wjwwood: Ah yeah, that makes sense.

PR #321, opened 2018-12-13
Use wall clock for Waitable tests.

hidmic:
This pull request attempts to fix failing `Waitable` tests as seen in https://ci.ros2.org/view/nightly/job/nightly_linux_release/1023/testReport/, https://ci.ros2.org/view/nightly/job/nightly_xenial_linux-aarch64_release/194/testReport/ and https://ci.ros2.org/view/nightly/job/nightly_xenial_linux_release/188/testReport/. I'm no longer able to reproduce the issue locally with this fix. 

My current hypothesis as to what's going is that the original test hooks up the timer to the default `Node` clock, which is of `RCL_ROS_TIME` type i.e. clock as published on a well known topic. Because of this, [the timer sets up a guard condition and waits for it to be triggered](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/timer.c#L155), but that never happens because for this test there's no clock topic being published in the first place. 

As to why it only fails for `Release` builds, it has to do with time. Until an `RCL_ROS_TIME` type clock becomes active, [it reports wall clock time](https://github.com/ros2/rcl/blob/master/rcl/src/rcl/time.c#L66). If the time it takes the code to reach the first `rmw_wait` call is long enough for the timer to expire, then it will not block because the first wait wakes up because of the `/parameter_events` subscription. If it's too fast, then the `Executor` will go into a second wait cycle from which it'll never wake up (because no one's going to trigger the timer's guard condition).

### DIFF
--- a/test_rclcpp/test/test_waitable.cpp
+++ b/test_rclcpp/test/test_waitable.cpp
@@ -92,7 +92,8 @@ public:
 TEST(CLASSNAME(test_waitable, RMW_IMPLEMENTATION), waitable_with_timer) {
   if (!rclcpp::ok()) {rclcpp::init(0, nullptr);}
   auto node = rclcpp::Node::make_shared("waitable_with_timer");
-  auto waitable = WaitableWithTimer::make_shared(node->get_clock());
+  auto wall_clock = rclcpp::Clock::make_shared(RCL_SYSTEM_TIME);
+  auto waitable = WaitableWithTimer::make_shared(wall_clock);
   auto group = node->create_callback_group(rclcpp::callback_group::CallbackGroupType::Reentrant);
   node->get_node_waitables_interface()->add_waitable(waitable, group);
 


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5895/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2440/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4886/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5799/)

Linux Release [Build Status](https://ci.ros2.org/job/ci_linux/5896/)
Linux Xenial Release [Build Status](https://ci.ros2.org/job/ci_linux/5897/)

dirk-thomas:
> the default Node clock, which is of `RCL_ROS_TIME` type i.e. clock as published on a well known topic.

That is incorrect. See http://design.ros2.org/articles/clock_and_time.html#ros-time Until the `use_sim_time` parameter is enables ROS time is the same as system time.

If this patch actually fixes the problem then there is an underlying problem in ROS time which needs to be fixed.

hidmic:
> That is incorrect. See http://design.ros2.org/articles/clock_and_time.html#ros-time Until the use_sim_time parameter is enables ROS time is the same as system time.

I see. I did not recall that design document, so I wrongly inferred `RCL_ROS_TIME` behavior from code. In any case, then the deeper problem is that the guard condition for a timer that's hooked to an `RCL_ROS_TIME` type clock is never triggered unless any of the `rcl_*_ros_time_override` functions are called that from rclcpp are in turn called from [`TimeSource`](https://github.com/ros2/rclcpp/blob/master/rclcpp/src/rclcpp/time_source.cpp#L160).

sloretz:
closing now that ros2/rcl#357 has been merged

PR #331, opened 2019-02-25
Add option to skip incompatible communication tests for address sanit…

bhatsach:
…izer

Introduce a new cmake option -DSKIP_ASAN_INCOMPATIBLE_TESTS to turn off
pub/sub and requester communication tests. This is needed to prevent the override
of LD_PRELOAD or library path to interfere from the working of address sanitizer

Connects to ros2/ci#245

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -149,6 +149,8 @@ if(BUILD_TESTING)
   endfunction()
 
   macro(multi_targets)
+    # Define option to skip tests which interfere with address sanitizer.
+    option(SKIP_ASAN_INCOMPATIBLE_TESTS "Skip tests which are address sanitizer incompatible" OFF)
     # test publish / subscribe messages
     if(rmw_implementation1 STREQUAL rmw_implementation2)
       set(suffix "${suffix}__${rmw_implementation1}")
@@ -168,6 +170,9 @@ if(BUILD_TESTING)
       set(rmw_implementation2_is_fastrtps TRUE)
     endif()
     if(WIN32 AND (rmw_implementation1_is_fastrtps OR rmw_implementation2_is_fastrtps))
+      set(SKIP_TEST "SKIP_TEST")
+    endif()
+    if(SKIP_ASAN_INCOMPATIBLE_TESTS)
       set(SKIP_TEST "SKIP_TEST")
     endif()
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Skipping all these tests sounds not like a good choice. It would be valuable t actually cover these cases.

bhatsach:
> Skipping all these tests sounds not like a good choice. It would be valuable t actually cover these cases.

Atleast for the current address sanitizer report, it does not seem to be adding extra value as the leaks detected by test_communication are already being detected by the other test_* and osrf_testing_tools tests.

dirk-thomas:
> Atleast for the current address sanitizer report, it does not seem to be adding extra value as the leaks detected by test_communication are already being detected by the other test_* packages

Then why include the `test_communication` package in the build in the first place?

bhatsach:
We could skip test_communication package entirely but I am more in favor of having this option based logic because:
* There are 16 leaks which are still being detected by the non skipped tests. I believe most/all of these are already being detected by other tests, but the more the tests it's better to detect the leaks
* When some other developer is trying to introduce address sanitizer or a similar feature to this package, he will not be starting from scratch and can use the fact that only some of the tests are incompatible

dirk-thomas:
> Atleast for the current address sanitizer report, it does not seem to be adding extra value as the leaks detected by test_communication are already being detected by the other test_* packages

Here you argue that disabling some test is fine because these cases are also covered elsewhere.

> I believe most/all of these are already being detected by other tests, but the more the tests it's better to detect the leaks

And here you argue that having additional tests being covered is good. These arguments are contradicting each other.

That is why I argued that it would make sense that the sanitizers cover all the tests (rather than a subset).

> that only some of the tests are incompatible

Maybe you can clarify what makes these tests "incompatible". I would think that it should be possible to use the sanitizers with these tests - it might just need some additional work?

bhatsach:
> > Atleast for the current address sanitizer report, it does not seem to be adding extra value as the leaks detected by test_communication are already being detected by the other test_* packages
> 
> Here you argue that disabling some test is fine because these cases are also covered elsewhere.
> 
> > I believe most/all of these are already being detected by other tests, but the more the tests it's better to detect the leaks
> 
> And here you argue that having additional tests being covered is good. These arguments are contradicting each other.
It's just that the extra tests can be easily run without any need for LD_PRELOAD or such changes, hence I prefer it to be there. 
> 
> That is why I argued that it would make sense that the sanitizers cover all the tests (rather than a subset).
> 
> > that only some of the tests are incompatible
> 
> Maybe you can clarify what makes these tests "incompatible". I would think that it should be possible to use the sanitizers with these tests - it might just need some additional work?

These tests are preventing the asan library from being the first in the library list and here's the exact error message:
"ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD"
The only way I know to fix this is to set the LD_PRELOAD environment variable to libasan path before triggering the test.

bhatsach:
Closing this PR as we are looking for a solution which will avoid skipping the tests and help address sanitizer work with the test_communication tests

PR #338, opened 2019-03-14
Prepare 0.6.1

nuclearsandwich:
This commit should be fast-forwarded onto the crystal branch rather than squashed.

Since we don't do bloom releases for system_tests our release discipline has lapsed. There's a 0.6.0 tag for crystal but the package.xml version was never updated. This jumps them from 0.4.0 to 0.6.1 to enable backporting of test fixes.

### DIFF
--- a/test_cli/package.xml
+++ b/test_cli/package.xml
@@ -2,7 +2,7 @@
 <?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="2">
   <name>test_cli</name>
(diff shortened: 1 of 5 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6412/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2795/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5270/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6244/)

PR #343, opened 2019-04-11
Change use of deprecated rclcpp APIs

emersonknapp:
connects to ros2/rclcpp#673

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -68,9 +68,11 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
     };
 
   // Try to saturate the MultithreadedExecutor's thread pool with subscriptions
+  rclcpp::SubscriptionOptions subscription_options;
+  subscription_options.callback_group = callback_group;
   for (uint32_t i = 0; i < num_messages; ++i) {
     auto sub = node->create_subscription<test_rclcpp::msg::UInt32>(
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -68,9 +68,11 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
     };
 
   // Try to saturate the MultithreadedExecutor's thread pool with subscriptions
+  rclcpp::SubscriptionOptions<> sub_opts;
```
wjwwood: Can you do something like this in `rclcpp`?

```c++
namespace rclcpp
{
  template<typename Allocator = std::allocator<void>>
  struct SubscriptionOptionsWithAllocator
  {
    // ...
  };

  using SusbscriptionOptions = SubscriptionOptionsWithAllocator<std::allocator<void>>;
}
```

In order to avoid the bare `<>`? We do something like this in our generated messages.

wjwwood: Also, please avoid abbreviation where you can, i.e. `subscription_options`. (it helps people for whom English is not their native language)

emersonknapp: De-abbreviated - Added the default specialization of SubscriptionOptions per suggestion. It makes the core spots more verbose but should be more straightforward for the majority end users, I think I like it.

Have updated all 3 of these PRs

PR #344, opened 2019-04-12
Update to be compatible with new rmw_qos_profile_t

mm318:
Some new fields are being added to rmw_qos_profile_t, and this package needs to be updated to be compatible with the newer definition of the struct.

Connects to ros2/rmw#173

### DIFF
--- a/test_rclcpp/test/test_client_scope_consistency_client.cpp
+++ b/test_rclcpp/test/test_client_scope_consistency_client.cpp
@@ -39,14 +39,7 @@ TEST(CLASSNAME(service_client, RMW_IMPLEMENTATION), client_scope_consistency_reg
   auto node = rclcpp::Node::make_shared("client_scope_consistency_regression_test");
 
   // Replicate the settings that caused https://github.com/ros2/system_tests/issues/153
-  rmw_qos_profile_t rmw_qos_profile =
-  {
-    RMW_QOS_POLICY_HISTORY_KEEP_LAST,
-    10,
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/test_client_scope_consistency_client.cpp
```diff
@@ -39,14 +39,7 @@ TEST(CLASSNAME(service_client, RMW_IMPLEMENTATION), client_scope_consistency_reg
...
   // Replicate the settings that caused https://github.com/ros2/system_tests/issues/153
-  rmw_qos_profile_t rmw_qos_profile =
-  {
-    RMW_QOS_POLICY_HISTORY_KEEP_LAST,
-    10,
-    RMW_QOS_POLICY_RELIABILITY_RELIABLE,
-    RMW_QOS_POLICY_DURABILITY_VOLATILE,
-    false
-  };
+  rmw_qos_profile_t rmw_qos_profile = rmw_qos_profile_default;
```
wjwwood: This doesn't look right, are the settings you're removing the exact same as what's in the default?

mm318: I double checked this one, and it looks like it is (https://github.com/ros2/rmw/pull/173/files#diff-3b257fc3e609df4f195297e5ee8fe258L43)

wjwwood: Ok, never mind, thanks.

PR #347, opened 2019-04-22
QoS System Tests

dabonnie:
Added basic system tests for the rclcpp QoS features

- liveliness
- deadline
- lifespan

Relates to ros2/design#212

### DIFF
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -8,7 +8,7 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-  add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic -Werror)
 endif()
 
 find_package(ament_cmake REQUIRED)
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -1,0 +1,53 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(test_quality_of_service)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+    set(CMAKE_CXX_STANDARD 14)
+endif()
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+
+if(BUILD_TESTING)
+
+    find_package(ament_lint_auto REQUIRED)
+    find_package(ament_cmake_pytest REQUIRED)
+    find_package(ament_cmake_gtest REQUIRED)
+    find_package(rclcpp REQUIRED)
+    find_package(rcutils)
+    find_package(rmw REQUIRED)
+    find_package(std_msgs REQUIRED)
+
+    # get the rmw implementations ahead of time
+    find_package(rmw_implementation_cmake REQUIRED)
+    get_available_rmw_implementations(rmw_implementations)
+
+    foreach(rmw_implementation ${rmw_implementations})
+        find_package("${rmw_implementation}" REQUIRED)
+    endforeach()
+
+    include_directories(include)
+    ament_lint_auto_find_test_dependencies()
+
+    # build a gtest executable with an implemented main
+    function(custom_gtest_executable target)
+        ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
+        target_compile_definitions(${target}
+                PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
+        ament_target_dependencies(${target}
+                "rclcpp",
+                "rcutils",
+                "std_msgs")
+        install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
+    endfunction()
+
+    custom_gtest_executable(test_liveliness "test/test_liveliness.cpp")
+
+endif()
+
+ament_package()
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -35,9 +35,13 @@ if(BUILD_TESTING)
 
   # build a gtest executable with an implemented main
   function(custom_gtest_executable target)
-    ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
+    ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp
+      test/publisher.cpp
+      test/qos_test_node.cpp
+      test/qos_utilities.cpp
+      test/subscriber.cpp)
     target_compile_definitions(${target}  PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
-    ament_target_dependencies(${target} "rclcpp" "rcutils" "std_msgs")
+    ament_target_dependencies(${target} rclcpp rcutils std_msgs)
     install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
   endfunction()
 
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -36,10 +36,10 @@ if(BUILD_TESTING)
   # build a gtest executable with an implemented main
   function(custom_gtest_executable target)
     ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp
-      test/publisher.cpp
       test/qos_test_node.cpp
-      test/qos_utilities.cpp
-      test/subscriber.cpp)
+      test/qos_test_publisher.cpp
+      test/qos_test_subscriber.cpp
+      test/qos_utilities.cpp)
     target_compile_definitions(${target}  PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
     ament_target_dependencies(${target} rclcpp rcutils std_msgs)
     install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -37,16 +37,13 @@ if(BUILD_TESTING)
   function(custom_gtest_executable target)
     ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
     target_compile_definitions(${target}  PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
-    ament_target_dependencies(${target}
-      "rclcpp",
-      "rcutils",
-      "std_msgs")
+    ament_target_dependencies(${target} "rclcpp" "rcutils" "std_msgs")
     install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
   endfunction()
 
-  custom_gtest_executable(test_liveliness "test/test_liveliness.cpp")
   custom_gtest_executable(test_deadline "test/test_deadline.cpp")
   custom_gtest_executable(test_lifespan "test/test_lifespan.cpp")
+  custom_gtest_executable(test_liveliness "test/test_liveliness.cpp")
 
 endif()
 
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -16,7 +16,6 @@ find_package(ament_cmake REQUIRED)
 if(BUILD_TESTING)
 
   find_package(ament_lint_auto REQUIRED)
-  find_package(ament_cmake_pytest REQUIRED)
   find_package(ament_cmake_gtest REQUIRED)
   find_package(rclcpp REQUIRED)
   find_package(rcutils)
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -4,49 +4,49 @@ project(test_quality_of_service)
 
 # Default to C++14
 if(NOT CMAKE_CXX_STANDARD)
-    set(CMAKE_CXX_STANDARD 14)
+  set(CMAKE_CXX_STANDARD 14)
 endif()
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    add_compile_options(-Wall -Wextra -Wpedantic)
+  add_compile_options(-Wall -Wextra -Wpedantic)
 endif()
 
 find_package(ament_cmake REQUIRED)
 
 if(BUILD_TESTING)
 
-    find_package(ament_lint_auto REQUIRED)
-    find_package(ament_cmake_pytest REQUIRED)
-    find_package(ament_cmake_gtest REQUIRED)
-    find_package(rclcpp REQUIRED)
-    find_package(rcutils)
-    find_package(rmw REQUIRED)
-    find_package(std_msgs REQUIRED)
+  find_package(ament_lint_auto REQUIRED)
+  find_package(ament_cmake_pytest REQUIRED)
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(rclcpp REQUIRED)
+  find_package(rcutils)
+  find_package(rmw REQUIRED)
+  find_package(std_msgs REQUIRED)
 
-    # get the rmw implementations ahead of time
-    find_package(rmw_implementation_cmake REQUIRED)
-    get_available_rmw_implementations(rmw_implementations)
+  # get the rmw implementations ahead of time
+  find_package(rmw_implementation_cmake REQUIRED)
+  get_available_rmw_implementations(rmw_implementations)
+  foreach(rmw_implementation ${rmw_implementations})
+    find_package("${rmw_implementation}" REQUIRED)
+  endforeach()
 
-    foreach(rmw_implementation ${rmw_implementations})
-        find_package("${rmw_implementation}" REQUIRED)
-    endforeach()
+  include_directories(include)
+  ament_lint_auto_find_test_dependencies()
 
-    include_directories(include)
-    ament_lint_auto_find_test_dependencies()
+  # build a gtest executable with an implemented main
+  function(custom_gtest_executable target)
+    ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
+    target_compile_definitions(${target}  PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
+    ament_target_dependencies(${target}
+    "rclcpp",
+    "rcutils",
+    "std_msgs")
+    install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
+  endfunction()
 
-    # build a gtest executable with an implemented main
-    function(custom_gtest_executable target)
-        ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
-        target_compile_definitions(${target}
-                PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
-        ament_target_dependencies(${target}
-                "rclcpp",
-                "rcutils",
-                "std_msgs")
-        install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
-    endfunction()
-
-    custom_gtest_executable(test_liveliness "test/test_liveliness.cpp")
+  custom_gtest_executable(test_liveliness "test/test_liveliness.cpp")
+  custom_gtest_executable(test_deadline "test/test_deadline.cpp")
+  custom_gtest_executable(test_lifespan "test/test_lifespan.cpp")
 
 endif()
 
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -38,9 +38,9 @@ if(BUILD_TESTING)
     ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
     target_compile_definitions(${target}  PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
     ament_target_dependencies(${target}
-    "rclcpp",
-    "rcutils",
-    "std_msgs")
+      "rclcpp",
+      "rcutils",
+      "std_msgs")
     install(TARGETS ${target} DESTINATION lib/${PROJECT_NAME})
   endfunction()
 
--- a/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
+++ b/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
@@ -39,12 +39,16 @@ class QosRclcppTestFixture : public ::testing::Test
 protected:
   void SetUp() override;
   void TearDown() override;
-  /// executor used to submit work
+  // executor used to submit work
   std::shared_ptr<rclcpp::executors::SingleThreadedExecutor> executor;
-  // empty promise
+  // empty promise used with dummy_future
   std::promise<bool> empty_promise;
-  // no future
+  // a dummy to make executor timeout
   std::shared_future<bool> dummy_future;
+  // test fixture publisher node
+  std::shared_ptr<QosTestPublisher> publisher;
+  // test fixture subscriber node
+  std::shared_ptr<QosTestSubscriber> subscriber;
 };
 
 #endif  // TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
--- a/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
+++ b/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
@@ -15,7 +15,6 @@
 #ifndef TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
 #define TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
 
-#include "gtest/gtest.h"
 
 #include <chrono>
 #include <memory>
@@ -23,10 +22,11 @@
 #include <string>
 #include <tuple>
 
+#include "gtest/gtest.h"
+
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/string.hpp"
 
-using namespace std::chrono_literals;
 
 /// Helper time conversion method
 /**
--- a/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
+++ b/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
@@ -12,12 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#ifndef TEST_QUALITY_OF_SERVICE_QOS_UTILITIES_HPP
-#define TEST_QUALITY_OF_SERVICE_QOS_UTILITIES_HPP
+#ifndef TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
+#define TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
+
+#include "gtest/gtest.h"
 
 #include <chrono>
 #include <memory>
+#include <mutex>
 #include <string>
+#include <tuple>
 
 #include "rclcpp/rclcpp.hpp"
 #include "std_msgs/msg/string.hpp"
@@ -29,42 +33,44 @@ using namespace std::chrono_literals;
  * @param milliseconds
  * @return tuple of milliseconds converted to <seconds, nanoseconds>
  */
-std::tuple<size_t, size_t> chrono_milliseconds_to_size_t(const std::chrono::milliseconds & milliseconds);
+std::tuple<size_t, size_t> chrono_milliseconds_to_size_t(
+  const std::chrono::milliseconds & milliseconds);
 
-/// Simple publisher class used system tests
+/// Simple publishing node used for system tests
 class Publisher : public rclcpp::Node
 {
 public:
-    Publisher(std::string name,
-      std::string topic,
-      rclcpp::PublisherOptions<> pub_options,
-      std::chrono::milliseconds publish_period,
-      unsigned int max_publish_count = 0);
+  Publisher(
+    std::string name,
+    std::string topic,
+    rclcpp::PublisherOptions<> pub_options,
+    std::chrono::milliseconds publish_period);
 
-    ~Publisher();
-    void start_publishing();
+  ~Publisher();
+  void start_publishing();
+  void stop_publishing();
+  void toggle_publishing();
 
-    /// name of this publisher
-    std::string name_;
-    /// topic name to publish
-    std::string topic_;
+  /// name of this publisher
+  std::string name_;
+  /// topic name to publish
+  std::string topic_;
 
-    /// the timer of this publisher
-    rclcpp::TimerBase::SharedPtr timer_ = nullptr;
-    /// actual publishing mechanism
-    rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_ = nullptr;
+  /// the timer of this publisher
+  rclcpp::TimerBase::SharedPtr timer_ = nullptr;
+  /// actual publishing mechanism
+  rclcpp::Publisher<std_msgs::msg::String>::SharedPtr publisher_ = nullptr;
 
-    rclcpp::PublisherOptions<> pub_options_;
+  rclcpp::PublisherOptions<> pub_options_;
 
-    std::chrono::milliseconds publish_period_;
+  std::chrono::milliseconds publish_period_;
 
-    /// number of messages published
-    int sent_message_count_;
-    /// max number of messages to publish
-    int max_publish_count_;
-
+  /// number of messages published
+  int sent_message_count_;
+  /// if true then currently has an active publishing timer
+  bool is_publishing = false;
+  std::recursive_mutex toggle_publish_mutex;
 };
-
 
 class Subscriber : public rclcpp::Node
 {
@@ -74,8 +80,14 @@ public:
     std::string topic,
     rclcpp::SubscriptionOptions<> sub_options);
 
+  ~Subscriber();
+  void start_listening();
+  void stop_listening();
+  void toggle_listening();
+
   int received_message_count_ = 0;
-  int qos_event_count_ = 0;
+  bool is_listening = false;
+  std::recursive_mutex toggle_subscriber_mutex;
 
 private:
   std::string name_;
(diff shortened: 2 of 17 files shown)


dabonnie:
@thomas-moulard - please run the following CI job:

- Gist file: https://gist.githubusercontent.com/dabonnie/0b7fd2d1ea08dc28af5db66bed5d9cc9/raw/1b792fa15d43e5590f3fed4095f1f65396ee3047/ros2.repos
- Additional BUILD flags: --packages-up-to test_quality_of_service
- Additional TEST flags: --packages-up-to test_quality_of_service
- CI Job: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6767/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3068/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5560/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6591/)

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+  toggle_publish_mutex.lock();
+  is_publishing = false;
+
+  if (timer_) {
+    timer_->cancel();
+  }
+  toggle_publish_mutex.unlock();
+}
+
+void Publisher::toggle_publishing()
```
thomas-moulard: This is why you need the recursive mutex and this is very fragile, instead, consider one of the following options:

A/ refactor the code as follow (of course for real code I'd move the timer_callback into it's own function, etc.):

```c++
void Publisher::toggle_publishing_impl(bool allow_start, bool allow_stop) {
  auto timer_callback =
    [this]() -> void {
      this->sent_message_count_++;
      auto message = std_msgs::msg::String();
      std::ostringstream ss;
      ss << this->name_ << ": testing" << this->sent_message_count_;
      message.data = ss.str();
      RCLCPP_INFO(this->get_logger(), "%s", ss.str().c_str());
      this->publisher_->publish(message);
    };
  auto timer = this->create_wall_timer(publish_period_, timer_callback);

  {
    std::scoped_lock lock(toggle_publish_mutex_);
    if (is_publishing && allow_stop) {
      return;
    }
    if (!is_publishing && allow_start) {
      return;
    }
    if (is_publishing) {
      timer_->cancel();
      timer_.reset();
    } else {
      timer_ = timer;
    }
    is_publishing = !is_publishing;
  }
}

void Publisher::toggle_publishing() {
   return toggle_publishing_impl(true /* allow_start */, true /* allow_stop */);
}

void Publisher::start_publishing() {
   return toggle_publishing_impl(true /* allow_start */, false /* allow_stop */);
}

void Publisher::start_publishing() {
   return toggle_publishing_impl(false /* allow_start */, true /* allow_stop */);
}
```

B/ rename this `try_toggle_publishing` and cope with the (potential) differences between `is_publishing` state when you read it in toggle and in start by returning a boolean indicating the operation failed.

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+  toggle_publish_mutex.lock();
+
+  if (is_publishing) {
+    return;
+  }
+
+  is_publishing = true;
+
+  auto timer_callback =
+    [this]() -> void {
```
thomas-moulard: This is very long for a lambda, consider making this lambda a private function of this class.

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+  toggle_publish_mutex.lock();
+  if (is_publishing) {
+    stop_publishing();
+  } else {
+    start_publishing();
+  }
+  toggle_publish_mutex.unlock();
+}
+
+Publisher::~Publisher()
```
thomas-moulard: Destructor should *never* contain code that can fail.

Is it possible for cancel() to fail? If yes, it should not be in the dtor.

I don't know the code enough but it seems to be a code smell you're using a code with a separate "cleanup" function (`timer_->cancel()`) but this difference is lost at this level of abstraction. Is timer cancelling on dtor? If not, do we know why?

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+{
+  if (timer_) {
+    timer_->cancel();
+  }
+  if (publisher_) {
+    publisher_.reset();
+  }
+}
+
+Subscriber::Subscriber(
```
thomas-moulard: Same comments apply here.

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+using namespace std::chrono_literals;
+
+/// Test Deadline with a single publishing node and single subscriber node
+TEST_F(TestSetup, test_deadline) {
+  int expected_number_of_pub_events = 5;
+  const std::chrono::milliseconds deadline_duration = 1s;
+  std::tuple<size_t, size_t> deadline_duration_tuple = chrono_milliseconds_to_size_t(
+    deadline_duration);
+  const std::chrono::milliseconds max_test_length = 10s;
+  const std::chrono::milliseconds publish_rate = deadline_duration / 5;
```
thomas-moulard: `deadline_duration / expected_number_of_pub_events`

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  // used for lambda capture
+  std::shared_ptr<rclcpp::executors::SingleThreadedExecutor> exec = executor;
+  int total_number_of_publisher_deadline_events = 0;
```
thomas-moulard: Sounds like all those int could be a struct.

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+// limitations under the License.
+
+#include <memory>
+#include <string>
+#include <tuple>
+
+#include "test_quality_of_service/qos_utilities.hpp"
+
+
+std::tuple<size_t, size_t> chrono_milliseconds_to_size_t(
```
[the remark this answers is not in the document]
thomas-moulard: By the, aren't you trying to re-invent `time_t` by chance? If yes, using the type directly could be better?

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+    return;
+  }
+
+  is_publishing = true;
+
+  auto timer_callback =
+    [this]() -> void {
+      auto message = std_msgs::msg::String();
+
+      message.data = this->name_ + ": testing " + std::to_string(this->sent_message_count_++);
```
thomas-moulard: Are we sure we don't need to lock sent_message_count here?

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+  rclcpp::SubscriptionOptions<> subscriber_options;
+  subscriber_options.qos_profile = qos_profile;
+  subscriber_options.event_callbacks.deadline_callback =
+    [&last_pub_count,
+    &total_number_of_subscriber_deadline_events](rclcpp::QOSDeadlineRequestedInfo & event) -> void
+    {
+      RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "QOSDeadlineRequestedInfo callback");
+      total_number_of_subscriber_deadline_events++;
+      // assert the correct value on a change
+      ASSERT_EQ(1, event.total_count_change);
```
thomas-moulard: EXPECT? https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+      kill_experiment_timer->cancel();
+    });
+
+  exec->add_node(subscriber);
+  exec->add_node(publisher);
+
+  subscriber->start_listening();
+  publisher->start_publishing();
+
+  exec->spin();  // blocking
```
thomas-moulard: See http://docs.ros2.org/beta2/api/rclcpp/index.html

using spin_some or spin_until_future_completes would drastically reduce the complexity of this test.

test_quality_of_service/test/test_lifespan.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+#include "std_msgs/msg/string.hpp"
+
+#include "test_quality_of_service/qos_utilities.hpp"
+
+
+using namespace std::chrono_literals;
+
+TEST_F(TestSetup, test_deadline) {
+  std::chrono::milliseconds lifespan_duration = 1000ms;
```
thomas-moulard: nit: 1s?

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+/// Test Deadline with a single publishing node and single subscriber node
+TEST_F(TestSetup, test_deadline) {
+  int expected_number_of_pub_events = 5;
+  const std::chrono::milliseconds deadline_duration = 1s;
+  std::tuple<size_t, size_t> deadline_duration_tuple = chrono_milliseconds_to_size_t(
+    deadline_duration);
+  const std::chrono::milliseconds max_test_length = 10s;
+  const std::chrono::milliseconds publish_rate = deadline_duration / 5;
+
+  // used for lambda capture
```
thomas-moulard: you could as well capture this in the lambda directly.

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+#include "std_msgs/msg/string.hpp"
+
+#include "test_quality_of_service/qos_utilities.hpp"
+
+
+using namespace std::chrono_literals;
+
+/// Test Deadline with a single publishing node and single subscriber node
+TEST_F(TestSetup, test_deadline) {
```
thomas-moulard: Why not get rid of all the locks and do:

```c++
TEST_F(...) {
  // init
  start_publishing();
  spin for 10 sec;
  stop_publishing();
  spin for 10 sec;
  check deadline
}
```
sounds way simplier than making this MT?

Also, I *believe* that sending messages without spinning is an easy way to miss deadlines:

```c++
set deadline to 1 second
send message
wait 5 seconds
spin until message is received somehow
check that we actually missed the deadline
```

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+  // setup publishing options and callback
+  rclcpp::PublisherOptions<> publisher_options;
+  publisher_options.qos_profile = qos_profile;
+  publisher_options.event_callbacks.deadline_callback =
+    [&last_sub_count,
+    &total_number_of_publisher_deadline_events](rclcpp::QOSDeadlineOfferedInfo & event) -> void
+    {
+      RCLCPP_DEBUG(rclcpp::get_logger("rclcpp"), "QOSDeadlineOfferedInfo callback");
+      total_number_of_publisher_deadline_events++;
```
thomas-moulard: In the test like in the classes, I don't think `++` w/o locking is safe here (?)

test_quality_of_service/test/test_deadline.cpp
thomas-moulard: Depending on how rclcpp is implemented, asserting there may make the whole test crash which is annoying. Switching a boolean we'd check in the main thread is a way to avoid this problem if it happens.

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,121 @@
...
+  std::string topic("test_deadline");
+
+  auto publisher = std::make_shared<Publisher>("publisher", topic, publisher_options, publish_rate);
+  auto subscriber = std::make_shared<Subscriber>("subscriber", topic, subscriber_options);
+
+  // toggle publishing on and off to force deadline events
+  rclcpp::TimerBase::SharedPtr kill_publisher_timer = nullptr;
+  kill_publisher_timer = subscriber->create_wall_timer(
+    max_test_length / expected_number_of_pub_events,
+    [&publisher]() -> void {
```
thomas-moulard: I would advise capturing by copy here. This code is susceptible to the following race:

```
1. publisher.reset()
2. Callback is called concurrently
3. publisher contains a random value => CRASH
```

If you want to keep this as a ref, you need a lock unless you can ensure that the publisher value will never be touched in the main thread before the timer is cancelled which is hard to achieve (we can get it right for this PR, but next change to this file will probably fail to remember this nuance).

test_quality_of_service/test/test_liveliness.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+#include "std_msgs/msg/string.hpp"
+
+#include "test_quality_of_service/qos_utilities.hpp"
+
+using namespace std::chrono_literals;
+
+
+/// Test Automatic Liveliness with a single publishing node and single subscriber node
+TEST_F(TestSetup, test_automatic_liveliness_changed) {
```
thomas-moulard: Same comments as for the other test.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
+#define TEST_QUALITY_OF_SERVICE__QOS_UTILITIES_HPP_
+
```
thomas-moulard: spurious empty line?

test_quality_of_service/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+  endforeach()
+
+  include_directories(include)
+  ament_lint_auto_find_test_dependencies()
+
+  # build a gtest executable with an implemented main
+  function(custom_gtest_executable target)
+    ament_add_gtest_executable(${target} ${ARGN} test/${target}.cpp test/qos_utilities.cpp)
+    target_compile_definitions(${target}  PUBLIC "RMW_IMPLEMENTATION=${rmw_implementation}")
+    ament_target_dependencies(${target} "rclcpp" "rcutils" "std_msgs")
```
thomas-moulard: you don't need to quote literals which cannot be expended to a value containing a space:

"foo" does not need to be quoted
${FOO}-test needs to be quoted because it may expend to abc def-test

There are a few other gotchas detailed here: https://stackoverflow.com/questions/35847655/when-should-i-quote-variables

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
dabonnie: Will fix, surprised the linters didn't complain.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include <string>
+#include <tuple>
+
+#include "gtest/gtest.h"
+
+#include "rclcpp/rclcpp.hpp"
+#include "std_msgs/msg/string.hpp"
+
+
+/// Helper time conversion method
```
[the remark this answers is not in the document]
dabonnie: I simply did this for readability, but for the case that we'll run system tests to verify builds (hopefully soon!) I'll split into separate headers.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+ */
+std::tuple<size_t, size_t> chrono_milliseconds_to_size_t(
+  const std::chrono::milliseconds & milliseconds);
+
+/// Simple publishing node used for system tests
+class Publisher : public rclcpp::Node
+{
+public:
+  Publisher(
+    std::string name,
```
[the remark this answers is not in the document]
dabonnie: Will do.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+    std::string topic,
+    rclcpp::PublisherOptions<> pub_options,
```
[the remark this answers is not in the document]
dabonnie: Will do.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+    std::chrono::milliseconds publish_period);
+
+  ~Publisher();
```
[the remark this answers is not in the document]
dabonnie: Good catch.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+  void start_publishing();
+  void stop_publishing();
+  /// switch publishing state
+  void toggle_publishing();
+
+  /// name of this publisher (Node)
```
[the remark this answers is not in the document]
dabonnie: Will do.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+  std::string name_;
+
+  /// topic name to publish
+  std::string topic_;
+
+  /// the timer of this publisher
+  rclcpp::TimerBase::SharedPtr timer_ = nullptr;
```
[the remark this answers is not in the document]
dabonnie: Will remove from header and initialize in constructor.

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+  std::string topic_;
+
+  rclcpp::Subscription<std_msgs::msg::String>::SharedPtr subscription_ = nullptr;
+
+  /// subscription options needed for QoS settings
+  rclcpp::SubscriptionOptions<> sub_options_;
+};
+
+/// Simple text fixture: standard bring-up and teardown for tests
+class TestSetup : public ::testing::Test
```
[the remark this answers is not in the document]
dabonnie: Something like QosRclcppTestFixture seems appropriate. The fun joys of variable naming!

test_quality_of_service/test/qos_utilities.cpp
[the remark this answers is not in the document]
dabonnie: > This function name is not great, it's not clear what means size_t there and what is this tuple. It would be probably easier for the reader to read:

Naming inspired by (https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/TimeUnit.html#toMillis(long)) as it's a simple conversion method.

> By the, aren't you trying to re-invent `time_t` by chance? If yes, using the type directly could be better?

Not trying to re-invent types: most of the time usage I've seen in rclcpp using chrono, whereas the QoS settings are using ctime. Without a function the conversion work was repeated, so this was simply to illustrate how a user could set the QoS time appropriately.

Will change the method name from _chrono_milliseconds_to_size_t_ to _convert_chrono_milliseconds_to_size_t_.

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include <string>
+#include <tuple>
+
+#include "test_quality_of_service/qos_utilities.hpp"
+
+
+std::tuple<size_t, size_t> chrono_milliseconds_to_size_t(
+  const std::chrono::milliseconds & milliseconds)
+{
+  size_t seconds = milliseconds.count() / 1000;
```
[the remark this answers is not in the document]
dabonnie: Ok

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+  size_t nanoseconds = (milliseconds.count() % 1000) * 1000000;
```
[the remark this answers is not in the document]
dabonnie: Ok

test_quality_of_service/test/test_deadline.cpp
thomas-moulard: Found some ref for that:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-value-capture

thomas-moulard: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconc-data-by-value

test_quality_of_service/test/qos_utilities.cpp
dabonnie: Will do.

test_quality_of_service/test/qos_utilities.cpp
dabonnie: If we cared about sequential publishing then yes, but I don't see the need for locking publishing nor incrementing the overall published count.

test_quality_of_service/test/qos_utilities.cpp
dabonnie: > Destructor should _never_ contain code that can fail.
> 
> Is it possible for cancel() to fail? If yes, it should not be in the dtor.
> 

Why not simply catch specific exceptions in the destructor when attempting to cleanup>? At that point, if anything actually failed there hasn't been checking to account for such and destruction will continue. Doing so when freeing resources at least attempts to correct a user's mistake.

```
Publisher::~Publisher()
  try {
    // cleanup after ourselves just in case someone didn't do it properly
    if (timer_) {
        timer_->cancel();
    }
  } catch (const runtime_error& error) {
    // could try something, but seems unlikely at this point
  }
```

> I don't know the code enough but it seems to be a code smell you're using a code with a separate "cleanup" function (`timer_->cancel()`) but this difference is lost at this level of abstraction. Is timer cancelling on dtor? If not, do we know why?

I'm not following. There is no separate cleanup, as it was all done in the destructor. The timer should have been cancelled earlier by the user or other method, but this at least makes a best effort to do so if left running. Make sense?

test_quality_of_service/test/test_deadline.cpp
dabonnie: I didn't have that experience when attempting to write these tests :-( and found it easier to use timers as implemented.

test_quality_of_service/test/test_deadline.cpp
dabonnie: Almost. It's split between the pub and sub callbacks. If it was more than the current inputs needed I'd agree, but seems overkill for the current situation.

test_quality_of_service/test/test_deadline.cpp
dabonnie: > EXPECT? https://testing.googleblog.com/2008/07/tott-expect-vs-assert.html

Assert was used explicitly. If the callback fails there's no reason to burn CPU cycles (at the end of the test experiment) when we know that the rest of the test code will fail. The intent here was simply to fail fast.

test_quality_of_service/test/test_deadline.cpp
dabonnie: Could use atomic incrementation via https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith

test_quality_of_service/test/test_deadline.cpp
dabonnie: A copy won't work as the publisher has to be explicitly reset in order to generate the liveliness event. Will have to use a lock.

test_quality_of_service/test/test_deadline.cpp
dabonnie: I've implemented using spin_until_future_completes: passing in a future that does nothing but using the previously defined timer period to enforce test length.

test_quality_of_service/test/test_lifespan.cpp
dabonnie: Ok

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+  std::chrono::milliseconds publish_period_;
+
+  /// number of messages published
+  int sent_message_count_ = 0;
+
+  /// if true then currently has an active publishing timer
+  bool is_publishing = false;
+
+  /// mutex used for starting, stopping, and toggling publishing
+  std::recursive_mutex toggle_publish_mutex;
```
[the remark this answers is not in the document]
dabonnie: I ended up with a major refactor given I was rewriting the same code in both the pub and sub nodes. I've used a common base class that follows the guidelines here (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md CP.22)

test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
```diff
@@ -0,0 +1,138 @@
...
+
+  /// subscription options needed for QoS settings
+  rclcpp::SubscriptionOptions<> sub_options_;
+};
+
+/// Simple text fixture: standard bring-up and teardown for tests
+class TestSetup : public ::testing::Test
+{
+protected:
+  void SetUp() override
```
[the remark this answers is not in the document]
dabonnie: Ok

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+  pub_options_(pub_options),
+  publish_period_(publish_period)
+{
+  RCLCPP_INFO(this->get_logger(), "created publisher %s %s\n", name.c_str(), topic.c_str());
+  publisher_ = this->create_publisher<std_msgs::msg::String>(topic_, pub_options_);
+}
+
+void Publisher::start_publishing()
+{
+  toggle_publish_mutex.lock();
```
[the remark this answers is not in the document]
dabonnie: I've simply made the callback a class method and fixed the issues described.

test_quality_of_service/test/qos_utilities.cpp
dabonnie: This should be resolved anyways by using an atomic int to store the count.

test_quality_of_service/test/qos_utilities.cpp
```diff
@@ -0,0 +1,158 @@
...
+      this->publisher_->publish(message);
+    };
+  timer_ = this->create_wall_timer(publish_period_, timer_callback);
+
+  toggle_publish_mutex.unlock();
+}
+
+void Publisher::stop_publishing()
+{
+  toggle_publish_mutex.lock();
```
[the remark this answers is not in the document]
dabonnie: I'm not sure I see the value of assert: are all system tests run in debug? If so then I agree, otherwise we should probably write tests to ensure that timer is in the state we expect.

test_quality_of_service/test/qos_test_node.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+bool QosTestNode::get_started()
+{
+  return started_;
+}
+
+void QosTestNode::toggle()
+{
+  std::unique_lock<std::recursive_mutex> ulock {toggle_mutex_};
```
thomas-moulard: std::unique_lock<std::recursive_mutex> ulock(toggle_mutex_);

test_quality_of_service/test/qos_test_publisher.cpp
```diff
@@ -0,0 +1,71 @@
...
+  publisher_options_(publisher_options),
+  publish_period_(publish_period),
+  timer_(nullptr),
+  is_publishing(false)
+{
+  RCLCPP_INFO(this->get_logger(), "created publisher %s %s\n", name.c_str(), topic.c_str());
+  publisher_ = this->create_publisher<std_msgs::msg::String>(topic_, publisher_options_);
+}
+
+QosTestPublisher::~QosTestPublisher()
```
thomas-moulard: You can remove and add `~QosTestPublisher() = default;`

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,167 @@
...
+  const std::chrono::milliseconds test_duration = 10500ms;
+  const std::tuple<size_t, size_t> deadline_duration_tuple = convert_chrono_milliseconds_to_size_t(
+    deadline_duration);
+  const int expected_number_of_deadline_callbacks = test_duration / deadline_duration;
+
+  int total_number_of_subscriber_deadline_events = 0;
+  int last_sub_count = 0;
+
+  rmw_qos_profile_t qos_profile = rmw_qos_profile_default;
+  qos_profile.deadline.sec = std::get<0>(deadline_duration_tuple);
```
thomas-moulard: `std::tie(qos_profile.deadline.sec, qos_profile.deadline.nsec) = deadline_duration_tuple`
may work

dabonnie: Yup! Implemented in latest changes.

test_quality_of_service/test/qos_test_node.cpp
dabonnie: Implemented in latest changes.

test_quality_of_service/test/qos_test_publisher.cpp
dabonnie: Implemented in latest changes.

test_quality_of_service/test/test_deadline.cpp
dabonnie: Implemented with an atomic integer.

mm318:
Can we move this to "In Progress" to "In Review"?

dirk-thomas:
@thomas-moulard Please run CI for this.

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7124/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3340/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5834/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6944/)
T

mm318:
@thomas-moulard - please run the following CI job:

- Gist file: https://gist.githubusercontent.com/mm318/b416cec6f0d2631037ee21112a6107a9/raw/326ef78fcf090696cd8894fc0e28421733bca107/ros2.repos
- Additional BUILD flags: --packages-up-to test_quality_of_service
- Additional TEST flags: --packages-up-to test_quality_of_service
- CI Job: ci_launcher

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7125/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3341/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5835/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6945/)

nburek:
It looks like this was failing because it was trying to find_package on pytest, which was not used and not included in the package.xml. 

@thomas-moulard Could you please restart the tests with the latest commit. Thank you

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7162/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3374/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5865/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6994/)

mm318:
@thomas-moulard can you please re-run the CI?

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7180/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3391/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5886/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7014/)

mm318:
@nuclearsandwich @dirk-thomas this should be ready to merge. Thanks!

dirk-thomas:
First this needs to be reviewed by someone. @wjwwood Can you maybe take a look at this?

test_quality_of_service/test/test_deadline.cpp
```diff
@@ -0,0 +1,166 @@
...
+  rclcpp::QoS qos_profile(10);
+  qos_profile.deadline(deadline_duration);
+
+  // setup subscription options and callback
+  rclcpp::SubscriptionOptions subscriber_options;
+  subscriber_options.event_callbacks.deadline_callback =
+    [&last_sub_count,
+    &total_number_of_subscriber_deadline_events](rclcpp::QOSDeadlineRequestedInfo & event) -> void
+    {
+      RCLCPP_INFO(rclcpp::get_logger("rclcpp"), "QOSDeadlineRequestedInfo callback");
```
wjwwood: Use `Node::get_logger()`. In this case, `subscriber->get_logger()` I think.

test_quality_of_service/test/test_liveliness.cpp
```diff
@@ -0,0 +1,129 @@
...
+  qos_profile.liveliness(RMW_QOS_POLICY_LIVELINESS_AUTOMATIC);
+  qos_profile.liveliness_lease_duration(lease_duration);
+
+  // subscription options
+  rclcpp::SubscriptionOptions subscriber_options;
+  // subscriber Liveliness callback event
+  subscriber_options.event_callbacks.liveliness_callback =
+    [&total_number_of_liveliness_events](rclcpp::QOSLivelinessChangedInfo & event) -> void
+    {
+      RCLCPP_INFO(rclcpp::get_logger("rclcpp"), "QOSLivelinessChangedInfo callback");
```
wjwwood: Same here, might just want to avoid `rclcpp::get_logger()` in all cases possible, it's kind of a "bad code smell".

test_quality_of_service/test/test_deadline.cpp
mm318: Hmm, this callback is being created before `subscriber`, so in order to use `subscriber->get_logger()`, there will need to be some non-trivial changes.

thomas-moulard: Move line 65-68 to line 42 and replace `rclcpp::get_logger("rclcpp")` by `subscriber->get_logger()`. You'll need to also capture `subscriber` by reference.

test_quality_of_service/test/test_liveliness.cpp
thomas-moulard: +1

test_quality_of_service/test/test_deadline.cpp
mm318: Okay, but this is quite confusing because `subscriber` and `subscriber_options` will be circularly referencing each other.

wjwwood: You don't have to do it, but if you're having to bend over backwards to do this, then maybe you should rethink how you're organizing the code. Maybe the class that inherits from node shouldn't create the subscription in the constructor but by a method call. That way you could create the node, create the callback referencing the node, and then pass the options to a method on the node that creates the subscription.

mm318: I'm becoming familiar with the code. I see a way to do it now. I'll update the PR when ready.

mm318:
@thomas-moulard can you please re-run the CI for this PR (https://github.com/ros2/system_tests/pull/347#issuecomment-493269717)?

thomas-moulard:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7188/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3398/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5894/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7024/)

test_quality_of_service/test/test_deadline.cpp
mm318: Done. Can you please re-review, @wjwwood?

mm318:
Hi @dirk-thomas and @wjwwood, this should be ready to merge. The instability of the macOS build is already being tracked at https://github.com/ros2/build_cop/issues/199.

wjwwood:
I didn't see any, but just to confirm, there are not any other pr's that need to be merged with this one right?

mm318:
This one is a standalone PR.

Thank you!

PR #348, opened 2019-04-29
Corrected publish calls with shared_ptr signature

ivanpauno:
Connects to ros2/rclcpp#709


dirk-thomas:
In several cases same comment as here: https://github.com/ros2/rclcpp/pull/709#discussion_r279500228

Issue #359, opened 2019-05-06
Bad WStrings encoding when publishing over Connext on Windows 


hidmic:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - From source
- Version or commit hash:
  - Current master at f65f852432090d1f2431d0cf496cbf4e2006922a
- DDS implementation:
  - RTI Connext
- Client library (if applicable):
  - rclcpp & rclpy

#### Steps to reproduce issue
Run `rclcpp`-to-`rclpy` pub/sub tests using Connext on Windows:

```
colcon test --event-handlers=console_direct+ --packages-select test_communication --ctest-args " -R" " ^test_publisher_subscriber__rclcpp__rclpy__rmw_connext_cpp"
```

#### Expected behavior
All tests pass.

#### Actual behavior
All test pass but it can be clearly seen that the subscriber fails to match incoming WString messages with the following traceback:

```
21: [test_subscriber-24]     raise RuntimeError('received unexpected message %r' % msg)
21: INFO:test_subscriber-24-stderr:[test_subscriber-24]     raise RuntimeError('received unexpected message %r' % msg)
21: --- Logging error ---
21: Traceback (most recent call last):
21:   File "C:\Python37\lib\logging\__init__.py", line 1037, in emit
21:     stream.write(msg + self.terminator)
21:   File "C:\Python37\lib\encodings\cp1252.py", line 19, in encode
21:     return codecs.charmap_encode(input,self.errors,encoding_table)[0]
21: UnicodeEncodeError: 'charmap' codec can't encode character '\u4e09' in position 136: character maps to <undefined>
21: Call stack:
21:   File "C:\Python37\lib\runpy.py", line 193, in _run_module_as_main
21:     "__main__", mod_spec)
21:   File "C:\Python37\lib\runpy.py", line 85, in _run_code
21:     exec(code, run_globals)
21:   File "C:\Users\Administrator\nuclearsandwich_ws\install\launch_testing\Lib\site-packages\launch_testing\launch_test.py", line 165, in <module>
21:     main()
21:   File "C:\Users\Administrator\nuclearsandwich_ws\install\launch_testing\Lib\site-packages\launch_testing\launch_test.py", line 142, in main
21:     results = runner.run()
21:   File "C:\Users\Administrator\nuclearsandwich_ws\install\launch_testing\Lib\site-packages\launch_testing\test_runner.py", line 231, in run
21:     results[run] = worker.run()
[...]

PR #363, opened 2019-05-09
Make test_subscriber_cpp always fail gracefully.

### LINKED ISSUE #360: Failing rclpy-to-rclcpp WString pub/sub over Connext on Windows
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - From source
- Version or commit hash:
  - Current master at f65f852432090d1f2431d0cf496cbf4e2006922a
- DDS implementation:
  - RTI Connext
- Client library (if applicable):
  - rclcpp & rclpy

#### Steps to reproduce issue
Run `rclpy`-to-`rclcpp` pub/sub tests using Connext on Windows:

```
colcon test --event-handlers=console_direct+ --packages-select test_communication --ctest-args " -R" " ^test_publisher_subscriber__rclpy__rclcpp__rmw_connext_cpp"
```

#### Expected behavior
All tests pass.

#### Actual behavior

Subscription to a WString publisher fails with:

```
18: [test_subscriber-24] r
18: INFO:test_subscriber-24-stderr:[test_subscriber-24] r
18: [test_subscriber-24] eceived message does not match any expected message
18: INFO:test_subscriber-24-stderr:[test_subscriber-24] eceived message does not match any expected message
18: INFO:test_publisher-23-stdout:[test_publisher-23]
18: [ERROR] [test_subscriber-24]: process has died [pid 3156, exit code 3221226505, cmd 'C:/Users/Administrator/nuclearsandwich_ws/build/test_communication/Release/test_subscriber_cpp.exe WStrings /test_time_21_23_20'].
18: ERROR:test_subscriber-24:process has died [pid 3156, exit code 3221226505, cmd 'C:/Users/Administrator/nuclearsandwich_ws/build/test_communication/Release/test_subscriber_cpp.exe WStrings /test_time_21_23_20'].
```

#### Additional information

This issue does not manifest itself on `rclcpp`-to-`rclcpp` nor `rclpy`-to-`rclpy` nor `rclcpp`-to-`rclpy` tests.

The process terminates almost silently. Looking at the `rclcpp` subscriber [implementation](https://github.com/ros2/system_tests/blob/master/test_communication/test/test_subscriber.cpp#L51), this suggests that a crash occurs during `rclcpp::shutdown`.

hidmic:
Fixes #360.

### DIFF
--- a/test_communication/test/test_subscriber.cpp
+++ b/test_communication/test/test_subscriber.cpp
@@ -49,8 +49,6 @@ rclcpp::SubscriptionBase::SharedPtr subscribe(
         ++index;
       }
       if (!known_message) {
-        fprintf(stderr, "received message does not match any expected message\n");
-        rclcpp::shutdown();
         throw std::runtime_error("received message does not match any expected message");
       }
 
@@ -141,7 +139,14 @@ int main(int argc, char ** argv)
     rclcpp::shutdown();
     return 1;
   }
-  rclcpp::spin(node);
+
+  try {
+    rclcpp::spin(node);
+  } catch (const std::exception & e) {
+    std::cerr << e.what() << std::endl;
+    rclcpp::shutdown();
+    return 1;
+  }
 
   auto end = std::chrono::steady_clock::now();
   std::chrono::duration<float> diff = (end - start);
--- a/test_communication/test/test_subscriber.cpp
+++ b/test_communication/test/test_subscriber.cpp
@@ -143,7 +143,7 @@ int main(int argc, char ** argv)
   try {
     rclcpp::spin(node);
   } catch (const std::exception & e) {
-    std::cerr << e.what() << std::endl;
+    fprintf(stderr, "%s\n", e.what());
     rclcpp::shutdown();
     return 1;
   }


hidmic:
Windows CI (building for Release and testing up to `test_communication`):

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6848/)

test_communication/test/test_subscriber.cpp
```diff
@@ -141,7 +139,14 @@ int main(int argc, char ** argv)
     rclcpp::shutdown();
     return 1;
   }
-  rclcpp::spin(node);
+
+  try {
+    rclcpp::spin(node);
+  } catch (const std::exception & e) {
+    std::cerr << e.what() << std::endl;
```
dirk-thomas: Please use `fprintf` here (see https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#c-specific).

test_communication/test/test_subscriber.cpp
```diff
@@ -141,7 +139,14 @@ int main(int argc, char ** argv)
     rclcpp::shutdown();
     return 1;
   }
-  rclcpp::spin(node);
+
+  try {
+    rclcpp::spin(node);
+  } catch (const std::exception & e) {
```
clalancette: Just to make sure I understand correctly: if we throw a `std::runtime_error` in the callback above, we won't get the corruption, but we'll fail the test, correct?

Assuming that is the case, then don't we need to also look into the root of the test failure?

hidmic: Yeap, the root of the test failure is #359.

clalancette: Ah, I see, I didn't understand before.  OK, then this is fine with me with @dirk-thomas comment fixed.  Thanks for the clarification.

test_communication/test/test_subscriber.cpp
hidmic: Uh, how's that `fprintf` prevents `stderr` interleaving?

dirk-thomas: Both operators are executed separately and have a chance of being interleaved by any other concurrent write to `stdout`. The `printf` creates a single string including the newline and writes them in a single call to `stdout`. While the latter technicall doesn't guarantee the operation to be atomic it practically makes it much much less likely to be interleaved.

hidmic: Thanks for the explanation, fixed in bbc82e9.

hidmic:
May I get an approval here? As to why CI is green, there's an issue with launch testing and exit codes on Windows that I need to fix. Long story short, I made the same mistake than the old launch testing did and process that exit with a 1 are taken as successful.

PR #366, opened 2019-05-15
Improve tests for parameters

jacobperron:
Related to https://github.com/ros2/rclcpp/issues/705

In particular, added tests for attempting to set undeclared parameters.

### DIFF
--- a/test_rclcpp/test/parameter_fixtures.hpp
+++ b/test_rclcpp/test/parameter_fixtures.hpp
@@ -27,23 +27,10 @@
 
 const double test_epsilon = 1e-6;
 
-void declare_test_parameters(std::shared_ptr<rclcpp::Node> node)
+std::vector<rclcpp::Parameter>
+get_test_parameters()
 {
-  node->declare_parameter("foo");
-  node->declare_parameter("bar");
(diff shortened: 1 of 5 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7098/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3320/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5814/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6918/)

jacobperron:
It looks like the test times out on macOS and Windows, but I noticed that the same timeout happened in the macOS nightly (https://github.com/ros2/build_cop/issues/198). It's possible the test is flaky and this addition may make it worse. It's probably why I needed to increase the shutdown timeout (which appears to fix things for Linux). @wjwwood If it is okay with you, I will merge this and we can follow up with looking into the macOS and Windows failures as part of https://github.com/ros2/build_cop/issues/198.

wjwwood:
Fine with me.

PR #371, opened 2019-06-04
[test_cli] don't fail the build if the test is missing

mikaelarguedas:
Unveiled by https://github.com/ros2/ros2/issues/722

### DIFF
--- a/test_cli/CMakeLists.txt
+++ b/test_cli/CMakeLists.txt
@@ -41,9 +41,11 @@ if(BUILD_TESTING)
     # https://github.com/ros2/system_tests/issues/365
     SKIP_TEST
   )
-  set_tests_properties(test_params_yaml
-    PROPERTIES DEPENDS
-      initial_params_rclcpp)
+  if(TEST test_params_yaml)
+    set_tests_properties(test_params_yaml
+      PROPERTIES DEPENDS
+        initial_params_rclcpp)
+  endif()
 endif()
 
 ament_auto_package()


test_cli/CMakeLists.txt
```diff
@@ -41,9 +41,11 @@ if(BUILD_TESTING)
     # https://github.com/ros2/system_tests/issues/365
     SKIP_TEST
   )
-  set_tests_properties(test_params_yaml
-    PROPERTIES DEPENDS
-      initial_params_rclcpp)
+  if(TEST test_params_yaml)
```
sloretz: What causes this target not to exist? If `BUILD_TESTING` is set, failing the build when a test is missing seems like the right behavior.

dirk-thomas: E.g. `pytest` not being available.

mikaelarguedas: > What causes this target not to exist?

pytest not being found prevents the target from being created: https://github.com/ros2/ros2/issues/722

>  If BUILD_TESTING is set, failing the build when a test is missing seems like the right behavior.

My understanding is that a warning should be provided but the package should not fail the build.
Similar patterns can be found in e.g. rclcpp and rcl:
https://github.com/ros2/rclcpp/blob/24080a458dd6cc227e29a97127209acca1ceb060/rclcpp/CMakeLists.txt#L149
https://github.com/ros2/rcl/blob/8085c6adbf30d76f8073632a4e2d2452aac366d9/rcl/test/cmake/rcl_add_custom_launch_test.cmake#L42

sloretz: > My understanding is that a warning should be provided but the package should not fail the build.
> Similar patterns can be found in e.g. rclcpp and rcl:

Oh. This looks like the case, but it doesn't seem right. With this pattern test coverage can be lost if a CI machine is configured incorrectly or the package.xml is missing a dependency, and the build will still pass.

dirk-thomas: We could certainly consider alternative behaviors in this case. But atm that is the behavior used throughout the code base. Since this patch fixes a configure failure I :+1: merging this.

dirk-thomas:
Thanks for the patch.

test_cli/CMakeLists.txt
sloretz: Opened an issue to continue discussion ros2/ros2#724

PR #372, opened 2019-06-21
Add tests for ros2action and ros2msg

ivanpauno:
The idea of this is start increasing testing coverage in all the `ros2cli` related command line tools, to avoid manual testing in the next release.

### DIFF
--- a/test_ros2cli/CMakeLists.txt
+++ b/test_ros2cli/CMakeLists.txt
@@ -19,34 +19,42 @@ if(BUILD_TESTING)
   find_package(ros_testing REQUIRED)
   find_package(rmw_implementation_cmake REQUIRED)
 
+  macro(custom_test_process_output test_name config_file_name)
+    set(TEST_NAME ${test_name})
+    set(TEST_CONFIGURATION_FILE ${config_file_name})
+    configure_file(
+      test/test_process_output_customizable.py.in
+      ${test_name}${target_suffix}.py.genexp
+      @ONLY
+    )
+    file(GENERATE
+      OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${test_name}${target_suffix}_$<CONFIG>.py"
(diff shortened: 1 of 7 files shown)


ivanpauno:
Added test for `ros2msg` in https://github.com/ros2/system_tests/pull/372/commits/af0fd2957d6373fd97a4eadc89a2de5fb90f6b44.
Refactored in https://github.com/ros2/system_tests/pull/372/commits/2149de3cf0a78714b4de216eaa029d8735509705 for avoiding duplicated code.

I will add test for the remaining verbs in separated PRs, after this one gets approved.

jacobperron:
Do these tests necessarily need to be in system_tests, or is it possible for them to live closer to the code being tested?

ivanpauno:
> Do these tests necessarily need to be in system_tests, or is it possible for them to live closer to the code being tested?

Yeah, maybe it's better. I added them here because these are more like "integration tests" than unit tests. But I think that that only justifies creating a new package (`test_ros2cli`), than moving them here.

If you agree, I will close this one and open a new PR in `ros2cli`.

jacobperron:
IMO it makes sense to move the package to ros2/ros2cli (similar to how launch_ros has [integration tests](https://github.com/ros2/launch_ros/tree/master/test_launch_ros) in the same repo). But, it would be good to get additional input from @ros2/team.

wjwwood:
I'm not sure why they were separated, but it may be due to these tests having a separate dependency on some message packages that otherwise would not be needed to build that repository. Personally, if it remains in a separate package, there's no reason that I know of to avoid having them in the same repository. But others might remember why it is this way better than me?

ivanpauno:
Closed in favor of https://github.com/ros2/ros2cli/pull/287.

mjcarroll:
@ivanpauno can we delete the branch?

PR #373, opened 2019-06-24
Fix some comparisons with a sign mismatch

cottsay:
Resolves build warnings due to `-Wsign-compare` with GCC 9.

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -75,7 +75,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
       node_topic_name, 5 * num_messages, callback, subscription_options);
     subscriptions.push_back(sub);
   }
-  ASSERT_TRUE(std::numeric_limits<int>::max() > subscriptions.size());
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -75,7 +75,7 @@ static inline void multi_consumer_pub_sub_test(bool intra_process)
       node_topic_name, 5 * num_messages, callback, subscription_options);
     subscriptions.push_back(sub);
   }
-  ASSERT_TRUE(std::numeric_limits<int>::max() > subscriptions.size());
+  ASSERT_TRUE(static_cast<size_t>(std::numeric_limits<int>::max()) > subscriptions.size());
```
sloretz: This doesn't seem like the right fix. Is `subscriptions.size()` limited to the size of an `int`? If not, maybe it should be `std::numeric_limits<size_t>::max()` instead?

sloretz: Ooops, what I posted makes no sense. subscriptions.size() can't be bigger than itself. nevermind

PR #375, opened 2019-07-11
Remove the test_cli dependency on launch.

clalancette:
It never uses it, so we don't need it.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

@sloretz It looks like you added this initially, but as far as I can tell this is never used in test_cli.  But let me know if I missed something.

### DIFF
--- a/test_cli/package.xml
+++ b/test_cli/package.xml
@@ -17,7 +17,6 @@
   <test_depend>ament_cmake_pytest</test_depend>
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7583/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3714/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7409/)

PR #322, opened 2018-12-14
Disable any cross-vendor communication tests for Fast-RTPS.

nuclearsandwich:
Builds are actually failing for all cross-vendor tests involving rmw_fastrtps_cpp not just those between Connext and Fast-RTPS.

PR #377, opened 2019-07-18
fix condition to not skip FastRTPS to FastRTPS pub/sub tests

dirk-thomas:
Fix regression from #322.

Before: https://ci.ros2.org/job/ci_windows/7438/testReport/(root)/test_communication/
After: https://ci.ros2.org/job/ci_windows/7439/testReport/(root)/test_communication/

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -166,8 +166,7 @@ if(BUILD_TESTING)
     endif()
     if(
       WIN32 AND
-      (rmw_implementation1_is_fastrtps OR rmw_implementation2_is_fastrtps) AND
-      NOT rmw_implementation1_is_fastrtps EQUAL rmw_implementation2_is_fastrtps
+      NOT "${rmw_implementation1_is_fastrtps}" STREQUAL "${rmw_implementation2_is_fastrtps}"
     )
       set(SKIP_TEST "SKIP_TEST")
     endif()
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -171,6 +171,11 @@ if(BUILD_TESTING)
       set(SKIP_TEST "SKIP_TEST")
     endif()
 
+    set(rmw_implementation2_is_connext FALSE)
+    if(rmw_implementation2 MATCHES "(.*)connext(.*)")
+      set(rmw_implementation2_is_connext TRUE)
+    endif()
+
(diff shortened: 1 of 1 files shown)


test_communication/CMakeLists.txt
```diff
@@ -164,7 +164,11 @@ if(BUILD_TESTING)
     if(rmw_implementation2 MATCHES "(.*)fastrtps(.*)")
       set(rmw_implementation2_is_fastrtps TRUE)
     endif()
-    if(WIN32 AND (rmw_implementation1_is_fastrtps OR rmw_implementation2_is_fastrtps))
+    if(
+      WIN32 AND
+      (rmw_implementation1_is_fastrtps OR rmw_implementation2_is_fastrtps) AND
+      NOT rmw_implementation1_is_fastrtps EQUAL rmw_implementation2_is_fastrtps
```
wjwwood: Why `rmw_implementation1_is_fastrtps EQUAL rmw_implementation2_is_fastrtps` versus `rmw_implementation1_is_fastrtps AND rmw_implementation2_is_fastrtps`?

dirk-thomas: I am not understanding your comment.

The added condition is `NOT a EQUAL b` which evaluates to false if both implementations are FastRTPS (or not FastRTPS which doesn't matter since the condition one line above is false in that case anyway).

dirk-thomas:
The XOR can be expresses as `NOT "${rmw1_isFastRTPS}" STREQUAL "{$rmw2_isFastRTPS}`. Done in 5ce5f1237e814258349ef898d480a06227225b0f.

Just using `NOT rmw1_isFastRTPS EQUAL rmw2_isFastRTPS` doesn't work - that is why the `STREQUAL` workaround is used in other places already: e.g. https://github.com/ros2/system_tests/blob/f1771f418288c6197d8c8b1d4e2b4ca780d3ab10/test_communication/CMakeLists.txt#L190

The additional condition `(rmw_implementation1_is_fastrtps OR rmw_implementation2_is_fastrtps)` is redundant and can be removed.

dirk-thomas:
This current patch successfully re-enables the FastRTPS-to-FastRTPS tests: see https://ci.ros2.org/job/ci_windows/7441/ (the build failed since it also uses ros2/test_interface_files#8).

dirk-thomas:
With the tests enabled the build fails (https://ci.ros2.org/job/ci_windows/7441/) due to a bug in launch: ros2/launch#280.

PR #380, opened 2019-08-02
[test_security] Generate security artifacts using sros2

mikaelarguedas:
Removes the need for expiring security artifacts committed to source tree (exception for an invalid certificate). Allows to test that the artifacts generated with sros2/master can be used successfully for secure communication.

Side changes bundled in this PR:
- Reenables xmllint test
- Fixes  `test_invalid_secure_node_creation_c.cpp` tests that were all testing the same code path due to environment modification leaked to following tests.
- Rename test cases
- print to stdout instead of stderr for info messages


test_security/CMakeLists.txt
```diff
@@ -335,6 +329,29 @@ if(BUILD_TESTING)
     # prevents finding it repeatedly in each local scope
     ament_find_gtest()
 
+    set(VALID_ROS_SECURITY_ROOT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/test/test_security_files")
+
+    # generate security artifacts using sros2
+    find_program(PROGRAM ros2)
```
jacobperron: The logic being added here looks similar to the existing cmake macro [`ros2_secure_node`](https://github.com/ros2/sros2/blob/b10a7e95a63692ab9632532dd8b8989e0be58fef/sros2_cmake/cmake/ros2_secure_node.cmake). What do you think about using that here instead?

mikaelarguedas: The two reasons I did not use `sros2_cmake` were:
- it didnt print the console output making it hard to know what was going on (fixed in https://github.com/ros2/sros2/pull/150)
- it doesn't provide a programmatic way to specify the keystore location (requires env var or defaults to install space). It could be worked around in multiple ways:
  - modify `sros2_cmake` interface to take a keystore argument:
    - ideally this would be a macro argument, as currently all the arguments are cmake arguments or environment variable, it may deserve a broader discussion about `sros2_cmake` API rather than just changing the keystore part of it
  - modify the environment before invoking sros2 and restoring it after
    - this works, but makes the code longer and it feels hacky to fiddle with the environment when the alternative is to just call the right function.

<details>

```diff
diff --git a/test_security/CMakeLists.txt b/test_security/CMakeLists.txt
index c7836ec..24bc841 100644
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -331,18 +331,22 @@ if(BUILD_TESTING)
 
     set(VALID_ROS_SECURITY_ROOT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/test/test_security_files")
 
-    # generate security artifacts using sros2
-    find_program(PROGRAM ros2)
-
-    set(node_names_list "/publisher;/subscriber;/publisher_missing_key;/publisher_invalid_cert")
-    set(generate_artifacts_command ${PROGRAM} security generate_artifacts -k ${VALID_ROS_SECURITY_ROOT_DIRECTORY} -n ${node_names_list})
-    execute_process(
-      COMMAND ${generate_artifacts_command}
-      RESULT_VARIABLE GENERATE_ARTIFACTS_RESULT
-      ERROR_VARIABLE GENERATE_ARTIFACTS_ERROR
+    find_package(sros2_cmake REQUIRED)
+    if(DEFINED ENV{ROS_SECURITY_ROOT_DIRECTORY})
+      set(PREVIOUS_ROS_SECURITY_ROOT_DIRECTORY $ENV{ROS_SECURITY_ROOT_DIRECTORY})
+    endif()
+    set(ENV{ROS_SECURITY_ROOT_DIRECTORY} "${VALID_ROS_SECURITY_ROOT_DIRECTORY}")
+    ros2_secure_node(
+      NODES
+      /publisher
+      /publisher_invalid_cert
+      /publisher_missing_key
+      /subscriber
     )
-    if(NOT ${POLICY_RESULT} EQUAL 0)
-      message(WARNING "Failed to generate security artifacts: ${GENERATE_ARTIFACTS_ERROR}")
+    if(DEFINED PREVIOUS_ROS_SECURITY_ROOT_DIRECTORY)
+      set(ENV{ROS_SECURITY_ROOT_DIRECTORY} ${PREVIOUS_ROS_SECURITY_ROOT_DIRECTORY})
+    else()
+      unset(ENV{ROS_SECURITY_ROOT_DIRECTORY})
     endif()
 
     # deleting key of /publisher_missing_key
```

</details>

So I sided with just calling `ros2 security generate_artifacts` directly for now until the API of sros2_cmake evolves into something more friendly for this use case.

jacobperron: :+1:
Updating the sros2_cmake API to accept the keystore as an argument makes sense to me.
In the meantime, this change LGTM as-is.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7856/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3919/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6410/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7716/)

PR #388, opened 2019-10-05
Don't set ignore_local_publications = true.

evshary:
According to [here](https://github.com/ros2/rclcpp/blob/master/rclcpp/include/rclcpp/intra_process_manager.hpp#L72), intra_process still needs local pub and sub to send special msg info.
Therefore, I think we shouldn't set ignore_local_publications to true.

Signed-off-by: evshary <evshary@gmail.com>

### DIFF
--- a/test_rclcpp/test/test_intra_process.cpp
+++ b/test_rclcpp/test/test_intra_process.cpp
@@ -62,10 +62,8 @@ TEST(CLASSNAME(test_intra_process_within_one_node, RMW_IMPLEMENTATION), nominal_
   rclcpp::executors::SingleThreadedExecutor executor;
 
   {
-    rclcpp::SubscriptionOptions options;
-    options.ignore_local_publications = true;
     auto subscriber = node->create_subscription<test_rclcpp::msg::UInt32>(
-      "test_intra_process", 10, callback, options);
+      "test_intra_process", 10, callback);
 
     // start condition
     ASSERT_EQ(0, counter);


dirk-thomas:
Can you please explain what motivates this change? Does the test pass / not pass for you before? Does it pass / not pass after?

evshary:
The PR tries to fix the CI issue [here](http://build.ros2.org/job/Eci__nightly-cyclonedds_ubuntu_bionic_amd64/lastCompletedBuild/testReport/test_rclcpp/test_intra_process_within_one_node__rmw_cyclonedds_cpp/nominal_usage/). I found that the issue appears after [this PR](https://github.com/ros2/rclcpp/pull/867).
After some studies, I think the reason is that `ignore_local_publications` is always false before, but the PR fixs it. (Refer to [here](https://github.com/ros2/rclcpp/pull/867/files#diff-66214cd6300588998f1883a79f8ce117L115)).
Back to the test, I can't think of any reason why `ignore_local_publications` should be true in the test, since intra_process should also rely on local pub.
It can pass in my local test after removing `ignore_local_publications`, so I suggest to remove the option.

hidmic:
CI (up to `test_rclcpp`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8416/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4327/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6851/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8315/)

hidmic:
Alright, we've got green lights. @evshary thank you for your contribution, keep them coming!

PR #393, opened 2019-11-19
Clean up bounded sequences check

rotu:
1. Use a macro to extract repeated logic
2. Check the sequence length so we get a test error message instead of segfaulting.
Signed-off-by: Dan Rose <dan@digilabs.io>

### DIFF
--- a/test_communication/test/test_messages_c.cpp
+++ b/test_communication/test/test_messages_c.cpp
@@ -1102,13 +1102,15 @@ void get_message(test_msgs__msg__BoundedSequences * msg, size_t msg_num)
   }
 }
 
-#define EXPECT_SEQUENCES_EQ(pred, a, b) \
+#define EXPECT_SEQUENCE_EQ(pairwise_check, a, b) \
   do { \
     EXPECT_EQ(a.size, b.size) << "Sequences have different length"; \
     for (size_t i = 0; i < std::min(a.size, b.size); ++i) { \
-      pred(a.data[i], b.data[i]) << "Sequences differ at element " << i; \
+      pairwise_check(a.data[i], b.data[i]) << "Sequences differ at element " << i; \
     } \
   } while (0)
+
+#define EXPECT_RCSTR_EQ(a, b) EXPECT_EQ(std::string(a.data, a.size), std::string(b.data, b.size))
 
 DEFINE_FINI_MESSAGE(test_msgs__msg__BoundedSequences)
 template<>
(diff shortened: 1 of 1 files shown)


test_communication/test/test_messages_c.cpp
```diff
@@ -1101,67 +1102,41 @@ void get_message(test_msgs__msg__BoundedSequences * msg, size_t msg_num)
...
+  EXPECT_SEQUENCES_EQ(EXPECT_EQ, expected_msg.int32_values, message.int32_values);
+  EXPECT_SEQUENCES_EQ(EXPECT_EQ, expected_msg.uint32_values, message.uint32_values);
+  EXPECT_SEQUENCES_EQ(EXPECT_EQ, expected_msg.int64_values, message.int64_values);
+  EXPECT_SEQUENCES_EQ(EXPECT_EQ, expected_msg.uint64_values, message.uint64_values);
+
+  EXPECT_EQ(expected_msg.string_values.size, message.string_values.size);
+  for (size_t i = 0; i < std::min(expected_msg.string_values.size, message.string_values.size);
+    ++i)
+  {
+    EXPECT_EQ(
```
hidmic: @rotu there's [`EXPECT_STREQ`](https://github.com/google/googletest/blob/master/googletest/docs/primer.md#string-comparison), I think we can afford an `EXPECT_STR_SEQUENCE_EQ` macro.

test_communication/test/test_messages_c.cpp
```diff
@@ -1101,67 +1102,41 @@ void get_message(test_msgs__msg__BoundedSequences * msg, size_t msg_num)
...
-  }
-  for (size_t i = 0; i < expected_msg.uint64_values.size; ++i) {
-    EXPECT_EQ(expected_msg.uint64_values.data[i],
-      message.uint64_values.data[i]);
-  }
-  for (size_t i = 0; i < expected_msg.string_values.size; ++i) {
-    EXPECT_EQ(0, strcmp(message.string_values.data[i].data,
-      expected_msg.string_values.data[i].data));
+
+  EXPECT_SEQUENCES_EQ(EXPECT_EQ, expected_msg.bool_values, message.bool_values);
```
hidmic: @rotu nit: syntax looks a bit awkward, can we add another set of macros so that it reads like:

```c++
EXPECT_SEQUENCE_EQ(/* ... */);
EXPECT_FLOAT_SEQUENCE_EQ(/* ... */);
EXPECT_DOUBLE_SEQUENCE_EQ(/* ... */);
```
?

test_communication/test/test_messages_c.cpp
rotu: Done. Except I switched to using std::string so that extra null terminators would be detected.

test_communication/test/test_messages_c.cpp
rotu: There are already too many macros. I felt dirty enough making a new one (vs a template, so the stack trace would be correct), but I would prefer to keep it as one.

test_communication/test/test_messages_c.cpp
```diff
@@ -1101,68 +1102,37 @@ void get_message(test_msgs__msg__BoundedSequences * msg, size_t msg_num)
...
 
+#define EXPECT_SEQUENCE_EQ(pairwise_check, a, b) \
+  do { \
+    EXPECT_EQ(a.size, b.size) << "Sequences have different length"; \
+    for (size_t i = 0; i < std::min(a.size, b.size); ++i) { \
+      pairwise_check(a.data[i], b.data[i]) << "Sequences differ at element " << i; \
+    } \
+  } while (0)
+
+#define EXPECT_RCSTR_EQ(a, b) EXPECT_EQ(std::string(a.data, a.size), std::string(b.data, b.size))
```
hidmic: @rotu what does `RCSTR` stands for? `rcutils` C string? Consider not using an abbreviation.

test_communication/test/test_messages_c.cpp
hidmic: > There are already too many macros. 

That's true, but at the same time I still find `EXPECT_SEQUENCES_EQ(EXPECT_EQ, ...)` a bit terse. That could easily be personal bias. @ivanpauno @dirk-thomas @shane what do you think? If you're OK with it, I'll approve.

hidmic:
Sorry for the delay @rotu !

test_communication/test/test_messages_c.cpp
ivanpauno: I prefer @hidmic style, but I don't mind strongly.

sloretz: I'm happy with it as is. I do like the sound of singular `EXPECT_SEQUENCE_EQ` since I read it as one sequence is expected to be equal to another.

If it was being added to a header to become a test utility, I would recommend an interface where `EXPECT_EQ` was the default if used like `EXPECT_SEQUENCE_EQ(seq1, seq2)`, but could be overridden with a third argument like `EXPECT_SEQUENCE_EQ(seq1, seq2, EXPECT_DOUBLE_EQ)`. However [it looks like it would be a lot more complex](https://stackoverflow.com/questions/3046889/optional-parameters-with-c-macros) than is needed for a macro used in a single file

test_communication/test/test_messages_c.cpp
rotu: ROS IDL-generated C STRING. What would you prefer it be called?

rotu: renamed

hidmic:
CI up to `system_tests`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4550/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7121/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8632/)

rotu:
woohoo!

hidmic:
Finally, Windows CI passed. Thank you for your contribution and patience @rotu!

PR #398, opened 2020-02-11
Specify stdout as the stream to look at.

clalancette:
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Connects to https://github.com/ros2/rcutils/pull/196


test_rclcpp/test/test_signal_handler_before_shutdown__expected_output.txt
```diff
@@ -1,7 +1,6 @@
 Registered custom signal handler.
 Called rclcpp::init.
 Waiting to give an opportunity for interrupt...
-[INFO] [rclcpp]: signal_handler(signal_value=2)
```
wjwwood: This line has some value in this test, imo... Maybe instead all of the messages being printed should also be using console logging.

clalancette: After thinking about it further, I agree.  The latest changes here restore it and change everything to using `RCLCPP_INFO`.

PR #401, opened 2020-03-05
Prevent subscriber tests from hanging forever if no work

rotu:
Signed-off-by: Dan Rose <dan@digilabs.io>

### DIFF
--- a/test_communication/test/subscriber_py.py
+++ b/test_communication/test/subscriber_py.py
@@ -62,7 +62,7 @@ def listener(message_name, namespace):
     spin_count = 1
     print('subscriber: beginning loop')
     while (rclpy.ok() and len(received_messages) != len(expected_msgs)):
-        rclpy.spin_once(node)
+        rclpy.spin_once(node, timeout_sec=0.1)
         spin_count += 1
         print('spin_count: ' + str(spin_count))
     node.destroy_node()


test_communication/test/subscriber_py.py
```diff
@@ -62,7 +62,7 @@ def listener(message_name, namespace):
     spin_count = 1
     print('subscriber: beginning loop')
     while (rclpy.ok() and len(received_messages) != len(expected_msgs)):
-        rclpy.spin_once(node)
+        rclpy.spin_once(node, timeout_sec=0)
```
dirk-thomas: I am not sure what the semantic of zero is here, but I would expect it to either return immediately or block forever which both seems not desirable.

rotu: The default `None` is block forever. 0 is execute immediately available work and return otherwise.

dirk-thomas: Makes sense since the new argument changes behavior. That means this change makes this a busy loop which isn't acceptable which will consume all of a CPU core.

I suggest to pass a timeout of e.g. `0.1s` instead.

rotu: Good point

dirk-thomas: Comparing this to `replier_py.py` maybe a larger timeout makes sense - to match the other test and avoid printing too frequently.

test_communication/test/subscriber_py.py
```diff
@@ -62,7 +62,7 @@ def listener(message_name, namespace):
     spin_count = 1
     print('subscriber: beginning loop')
     while (rclpy.ok() and len(received_messages) != len(expected_msgs)):
```
ivanpauno: This condition will still not be met, regardless of `spin_once` timeout, and this will still be waiting forever (or to the test to timeout). So, changing `spin_once` behavior from blocking to anything else just makes the loop more busy.

I think the original test case was ok, we can change the timeout of the test to a smaller value to avoid CI taking much longer when some of these tests fail.

claireyywang:
@rotu friendly ping

rotu:
Thanks! Yeah, I’m closing this due to user error. I had inferred from the test output that the listener having to be terminated was the problem. Opened #404.

PR #409, opened 2020-04-05
Make Connext use Connext's openssl on all platforms

mikaelarguedas:
Ubuntu Focal, Homebrew and chocolatey now provide OpenSSL 1.1.1
RTI Connext 5.3.1 supports only 1.0.2 so we need to modify the library path and the path for connext's version of openssl be used when running tests with connext. Use system OpenSSL for other rmw implementations

related to https://github.com/ros2/ci/pull/421

### DIFF
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -72,8 +72,6 @@ if(BUILD_TESTING)
       RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation}
       RMW_IMPLEMENTATION=${rmw_implementation}
       ROS_SECURITY_KEYSTORE=${KEYSTORE_DIRECTORY_NATIVE_PATH}
-      # on bionic, Connext needs RTI's OpenSSL binaries to be on the PATH
-      # at runtime as the system version of OpenSSL is not supported
       PATH="${TEST_PATH}"
     )
     if(TARGET ${target}${target_suffix})
@@ -159,8 +157,6 @@ if(BUILD_TESTING)
             "${CMAKE_CURRENT_BINARY_DIR}/test_secure_publisher_subscriber${test_suffix}_$<CONFIG>.py"
             TARGET test_secure_publisher_subscriber${test_suffix}
             APPEND_LIBRARY_DIRS "${append_library_dirs}"
-            # on bionic, Connext needs RTI's OpenSSL binaries to be on the PATH
-            # at runtime as the system version of OpenSSL is not supported
             ENV
             PATH="${TEST_PATH}"
             TIMEOUT 20
@@ -201,8 +197,6 @@ if(BUILD_TESTING)
             "${CMAKE_CURRENT_BINARY_DIR}/test_secure_publisher_subscriber${test_suffix}_$<CONFIG>.py"
             TARGET test_secure_publisher_subscriber${test_suffix}
             APPEND_LIBRARY_DIRS "${append_library_dirs}"
-            # on bionic, Connext needs RTI's OpenSSL binaries to be on the PATH
-            # at runtime as the system version of OpenSSL is not supported
             ENV
             PATH="${TEST_PATH}"
             TIMEOUT 20
@@ -243,8 +237,6 @@ if(BUILD_TESTING)
             "${CMAKE_CURRENT_BINARY_DIR}/test_secure_publisher_subscriber${test_suffix}_$<CONFIG>.py"
             TARGET test_secure_publisher_subscriber${test_suffix}
             APPEND_LIBRARY_DIRS "${append_library_dirs}"
-            # on bionic, Connext needs RTI's OpenSSL binaries to be on the PATH
-            # at runtime as the system version of OpenSSL is not supported
             ENV
             PATH="${TEST_PATH}"
             TIMEOUT 20
@@ -265,15 +257,11 @@ if(BUILD_TESTING)
     file(TO_CMAKE_PATH "${ENV_PATH}" ENV_PATH)
     set(TEST_PATH "${ENV_PATH}")
     if(rmw_implementation STREQUAL "rmw_connext_cpp")
-      if(UNIX AND NOT APPLE)
-        # TODO(mikaelarguedas) remove this condition the day we want to use RTI's OpenSSL libraries
-        #  on all platforms
-        # on bionic, Connext needs RTI's OpenSSL binaries to be on the PATH
-        # at runtime as the system version of OpenSSL is not supported
-        set(RTI_BIN_PATH "$ENV{RTI_OPENSSL_BIN}")
-        file(TO_CMAKE_PATH "${RTI_BIN_PATH}" RTI_BIN_PATH)
-        set(TEST_PATH "${RTI_BIN_PATH};${ENV_PATH}")
-      endif()
+      # Connext 5.3.1 needs RTI's OpenSSL binaries (based on EOL 1.0.2) to be
+      # on the PATH at runtime as the system version of OpenSSL is not supported
+      set(RTI_BIN_PATH "$ENV{RTI_OPENSSL_BIN}")
+      file(TO_CMAKE_PATH "${RTI_BIN_PATH}" RTI_BIN_PATH)
+      set(TEST_PATH "${RTI_BIN_PATH};${ENV_PATH}")
     endif()
 
     if(NOT WIN32)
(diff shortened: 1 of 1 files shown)


cottsay:
Could you please create a PR to update the ROS 2 installation instructions to set the appropriate environment variables to make this change work?

mikaelarguedas:
I'm not sure I got what you are referring to, could you please clarify which instructions you're thinking of?

I don't see the "normal" installation instructions detailing how to setup the security plugins for Connext. Are you referring to this page https://github.com/ros2/ros2_documentation/blob/master/source/Installation/Install-Connext-Security-Plugins.rst ? (noting that that page is not distro or platform specific)

Also the `RTI_OPENSSL_*` are made-up variables used only on CI and in this package. So they may not belong on official instructions

mikaelarguedas:
If the documentation update is not a blocker it would be awesome to get this PR reviewed / merged to fix ci on MacOS and hopefully unblock https://github.com/ros2/ci/pull/421

cottsay:
Special goo for our CI system doesn't give me the warm fuzzies :slightly_frowning_face: but since we're just making the existing behavior consistent across platforms, this seems like an OK change to make.

I'd was looking for documentation so that if/when we spin-up additional macOS workers on ci.ros2.org, we make sure that they get the older `openssl` installed and get the magic variables set. If it isn't documented anywhere, it's probably not going to happen.

@mjcarroll, did you follow any directions besides the official ROS 2 documentation when you set up `lore`?

nuclearsandwich:
> @mjcarroll, did you follow any directions besides the official ROS 2 documentation when you set up `lore`?

Lore was being shared between the OSRF and ROS 2 CI farms so it had additional setup performed for that. I've undone some of the Jenkins-related things that were done but not any other operations, which I also don't have an enumeration of.

nuclearsandwich:
> Also the `RTI_OPENSSL_*` are made-up variables used only on CI and in this package. So they may not belong on official instructions

If we have to fabricate variables for our CI to work we should at least share what we needed to do to get it working. Even if we don't tell people to replicate our variables, explaining what function they serve and how to achieve the same effect if they want to use connext security should IMO be documented

nuclearsandwich:
I talked about this PR with @cottsay a little bit and my least favorite thing about it is that it's being implemented "at runtime" in the tests rather than something we can configure as behavior whenever security is being used with connext. I can't think of a way to pin this behavior at the rmw_connext level that isn't worse than the problem it solves.

test_security/CMakeLists.txt
```diff
@@ -301,29 +289,28 @@ if(BUILD_TESTING)
...
+    #     OUTPUT_VARIABLE _out_var
+    #     ERROR_VARIABLE _error_var
+    #   )
+    #   if(NOT _retcode EQUAL 0)
+    #     message(FATAL_ERROR "command 'brew --prefix openssl' failed with error code '${_retcode}' and error message '${_error_var}'")
+    #   endif()
+    #   string(STRIP "${_out_var}" _out_var)
+    #   set(openssl_lib_path "${_out_var}/lib")
+    #   file(TO_NATIVE_PATH "${openssl_lib_path}" openssl_lib_path)
+    #   set(append_library_dirs "${append_library_dirs};${openssl_lib_path}")
```
nuclearsandwich, line 313: Are you keeping the above block commented so that it goes back into use when Connext is updated?

test_security/CMakeLists.txt
```diff
@@ -243,8 +237,6 @@ if(BUILD_TESTING)
             "${CMAKE_CURRENT_BINARY_DIR}/test_secure_publisher_subscriber${test_suffix}_$<CONFIG>.py"
             TARGET test_secure_publisher_subscriber${test_suffix}
             APPEND_LIBRARY_DIRS "${append_library_dirs}"
-            # on bionic, Connext needs RTI's OpenSSL binaries to be on the PATH
-            # at runtime as the system version of OpenSSL is not supported
```
nuclearsandwich, line 247: Went back and forth a bit about whether I thought we should keep some version of the comment here but it's a lot of duplication and I think the comment is better kept in the one place where we set the test path. :+1:

test_security/CMakeLists.txt
nuclearsandwich: https://github.com/ros2/system_tests/pull/409/files/593d7003e929be6a36c8538aaf0eabca13c450a9#diff-6fc729fbfd434d3b129f70533aa0346dR295-R296

yes. Hello it is Friday afternoon.

nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10082/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5669/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8202/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9985/)

test_security/CMakeLists.txt
mikaelarguedas: That was my feeling as well when making that change :+1: probably should've formilated that when I moved this PR from draft to ready

mikaelarguedas:
> If it isn't documented anywhere, it's probably not going to happen.

True

> Even if we don't tell people to replicate our variables, explaining what function they serve and how to achieve the same effect if they want to use connext security should IMO be documented

Yeah for a bit more history, it is needed on CI only because we want to use that specific version only for Connext but want to be able to use the system one `1.1.1*` for the other implementations.
Another reason the code (now commented) had to happen at runtime was because MacOS with SIP stripped the DYLD_LIBRARY_PATH, so setting that ahead of time had no effect.

I've been refraining from changing  https://github.com/ros2/ros2_documentation/blob/master/source/Installation/Install-Connext-Security-Plugins.rst that is written + maintained by RTI. That page does already specify to install their version of OpenSSL.

Best way forward is still to use a version of Connext that works with OpenSSL 1.1.1* (like Connext 6.0.1) ;) and remove all custom workarounds

mikaelarguedas:
For this change to fix CI, these `RTI_OPENSSL_*` variables need to be set on all macOS machines. Looks like the one used for testing this didnt have them set so the connext security tests failed

mikaelarguedas:
@cottsay @nuclearsandwich friendly :bellhop_bell: is there anything we can do to move this forward ?

There are currently no rmw implementation with passing security tests on MacOS, so this is challenging for validating new features before the freeze

nuclearsandwich:
> @cottsay @nuclearsandwich friendly bellhop_bell is there anything we can do to move this forward ?

I set up mini2 with Connext's openssl 1.0.2n installed to `/Applications/rti_connext_5.3.1/openssl-1.0.2n/...`

I posted [this branch](https://github.com/ros2/ci/tree/macos-rti-openssl) to ros2/ci which sets the RTI_OPENSSL_BIN and RTI_OPENSSL_LIBS environment variables to match the installation on mini2 and ran https://ci.ros2.org/job/test_ci_osx/308/ on mini2. With 128 security test failures it seems like either I did something wrong or there's an issue with the patch. I haven't had a chance to investigate further.

nuclearsandwich:
I'd love to address these test failures as it would cut down on the macos noise A LOT. I just don't have the personal bandwidth atm to do much but change configs and run tests. Feel free to take that CI branch and iterate on it if need be.

mikaelarguedas:
no code changes just a rebase.
Corresponding CI PR and jobs at https://github.com/ros2/ci/pull/436

sloretz:
@mikaelarguedas thanks for figuring this out. Is this then what needs to happen?

* Make sure lore, mini1, mini2, and mini3 have environment variables  `RTI_OPENSSL_BIN` and `RTI_OPENSSL_LIBS` set to the 1.0.2 openssl version in some global way
  * It sounds like they already have this based on the passing build badges in the description of ros2/ci#436
* Investigate why the connext security test failures still happened on mini3
* Merge ros2/ci#436 which sets these variables if they're not set
* Merge this PR

If that's correct, it sounds like We could merge this and ros2/ci#426 now since it resolves the failures on most of the machines, and that's better than the current state of test failures on all of the machines. Then mini3 can be investigated separately. How does that sound?

mikaelarguedas:
This PR had an unwanted side effect on windows which is now fixed at https://github.com/ros2/system_tests/pull/409/commits/28c4488c3f9e13fb4ec713593625ae232a50eed3.

See https://ci.ros2.org/job/ci_windows/10422/ using this change (windows security tests succeed)

> Is this then what needs to happen?

Yes as far as MacOS is concerned.
Ideally the MacOS machines would also be updated to use the latest openssl from brew before the release

For windows:

- Update windows to use the latest openssl: https://github.com/ros2/ci/pull/421
- Figure out why the tests are failing https://ci.ros2.org/job/ci_windows/10424

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10601/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6043/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8633/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10498/)

mikaelarguedas:
> Assuming that comes back green, can I just merge this? Does it have any other dependencies?

Merging as is should have no impact (what already fails will keep failing, what passes will keep passing) but give us the groundwork to fix things on top.
For this to improve things on macos you'll need https://github.com/ros2/ci/pull/436

clalancette:
aarch64 is expected to fail for connext-only.  macOS is already failing the builds that are failing.  Thus I'll merge this one.

nuclearsandwich:
Thanks @mikaelarguedas, and @clalancette for getting this over the edge.

PR #410, opened 2020-04-07
Use keystore root as security root directory, and not contexts folder

ivanpauno:
See https://github.com/ros2/rcl/pull/607.
Both should be merged together.

### DIFF
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -331,7 +331,7 @@ if(BUILD_TESTING)
     ament_find_gtest()
 
     set(KEYSTORE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/test/test_security_files")
-    if (WIN32)
+    if(WIN32)
       string(REPLACE "/" "\\\\" KEYSTORE_DIRECTORY_NATIVE_PATH "${KEYSTORE_DIRECTORY}")
     else()
       set(KEYSTORE_DIRECTORY_NATIVE_PATH "${KEYSTORE_DIRECTORY}")
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -71,7 +71,7 @@ if(BUILD_TESTING)
       ENV
       RCL_ASSERT_RMW_ID_MATCHES=${rmw_implementation}
       RMW_IMPLEMENTATION=${rmw_implementation}
-      ROS_SECURITY_ROOT_DIRECTORY=${VALID_ROS_SECURITY_ROOT_DIRECTORY}
+      ROS_SECURITY_ROOT_DIRECTORY=${KEYSTORE_DIRECTORY}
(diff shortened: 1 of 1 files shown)


ivanpauno:
I cannot reproduce the problem in my Windows VM, I will try to figure it out ...

ivanpauno:
> I cannot reproduce the problem in my Windows VM, I will try to figure it out ...

I forgot `-DSECURITY=ON` :man_facepalming: .

ivanpauno:
CI: https://github.com/ros2/rcl/pull/607#issuecomment-610608586 and https://github.com/ros2/rcl/pull/607#issuecomment-610952324.

mikaelarguedas:
Where the linux and MacOS tests ran with that change too ? (the path modification seem to happen unconditionally / on all platforms)

ivanpauno:
> Where the linux and MacOS tests ran with that change too ? (the path modification seem to happen unconditionally / on all platforms)

Thank you!
New CI jobs here https://github.com/ros2/rcl/pull/607#issuecomment-611592228.

test_security/CMakeLists.txt
```diff
@@ -331,7 +331,11 @@ if(BUILD_TESTING)
     ament_find_gtest()
 
     set(KEYSTORE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/test/test_security_files")
-    string(REPLACE "/" "\\\\" KEYSTORE_DIRECTORY_NATIVE_PATH "${KEYSTORE_DIRECTORY}")
+    if (WIN32)
```
mikaelarguedas, line 334:
```suggestion
    if(WIN32)
```

test_security/CMakeLists.txt
```diff
@@ -331,13 +331,17 @@ if(BUILD_TESTING)
     ament_find_gtest()
 
     set(KEYSTORE_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/test/test_security_files")
-    set(VALID_ROS_SECURITY_ROOT_DIRECTORY "${KEYSTORE_DIRECTORY}/contexts")
+    if(WIN32)
+      string(REPLACE "/" "\\\\" KEYSTORE_DIRECTORY_NATIVE_PATH "${KEYSTORE_DIRECTORY}")
```
mikaelarguedas, line 335: Nit^10: this could be done in place removing the need for a new variable and an else cause here

ivanpauno: I'm not sure if this https://github.com/ros2/system_tests/blob/0ec293e4bd6efc7bf935b9e2885a6bac35856e76/test_security/CMakeLists.txt#L355

will continue working well, as it's a mixture of a cmake and a "native" path.

I'm going to left this as-is, as I don't want to come back to Windows to test.

ivanpauno:
> Do you want to merge this and we'll do a follow-up for the enclave renaming ?

Yeah, the renaming should be addressed in a separate PR.

PR #421, opened 2020-04-13
disabled Connext-CycloneDDS WString tests

dirk-thomas:
Only testing macOS since the patch just removes some tests (independent of the platform):
* Before [Build Status](https://ci.ros2.org/job/ci_osx/8235/)
* After: [Build Status](https://ci.ros2.org/job/ci_osx/8238/)


dirk-thomas:
Since upstream is not planning to work on this in the near future (https://github.com/ros2/rmw_cyclonedds/issues/43#issuecomment-612286625) we should go ahead and disable the tests for now.

PR #423, opened 2020-04-14
updates since changes to message_info in rclcpp

wjwwood:
Follow up of https://github.com/ros2/rclcpp/pull/1047, which broke nightly.

### DIFF
--- a/test_rclcpp/test/test_avoid_ros_namespace_conventions_qos.cpp
+++ b/test_rclcpp/test/test_avoid_ros_namespace_conventions_qos.cpp
@@ -44,11 +44,11 @@ TEST(
   int counter = 0;
   auto callback =
     [&counter](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
-      const rmw_message_info_t & info) -> void
+      const rclcpp::MessageInfo & info) -> void
     {
       ++counter;
       printf("  callback() %d with message data %u\n", counter, msg->data);
-      ASSERT_FALSE(info.from_intra_process);
+      ASSERT_FALSE(info.get_rmw_message_info().from_intra_process);
     };
   auto qos = rclcpp::QoS(rclcpp::KeepLast(10)).avoid_ros_namespace_conventions(true);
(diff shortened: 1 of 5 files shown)


test_rclcpp/test/test_avoid_ros_namespace_conventions_qos.cpp
```diff
@@ -44,11 +44,11 @@ TEST(
   int counter = 0;
   auto callback =
     [&counter](test_rclcpp::msg::UInt32::ConstSharedPtr msg,
-      const rmw_message_info_t & info) -> void
+      const rclcpp::MessageInfo & info) -> void
```
ivanpauno, line 47: Is the previous callback signature still available?

This change is fine, but I'm just wondering if users that update to Foxy will see a warning being log or a build failure.

wjwwood: No. It is a build failure. I'm going to add a note to the migration guide for Foxy.

ivanpauno: Ok, I'm not sure if deprecation is possible or not, but if not adding a note to the migration notes sounds fine.

wjwwood: See: https://github.com/ros2/ros2_documentation/pull/606

wjwwood: I can work on trying to support the old signature, but I thought it was a) uncommonly used and b) made the signature resolution a bit trickier since it changes the type but now produces multiple signatures with the same arity.

ivanpauno: > a) uncommonly used

Yeah, that's true.

b) made the signature resolution a bit trickier since it changes the type but now produces multiple signatures with the same arity.

Ok, sounds good.

PR #430, opened 2020-05-01
Make `test_executor.spin_some_max_duration` more reliable

ivanpauno:
Fixes https://ci.ros2.org/view/nightly/job/nightly_win_rel/1537/testReport/junit/test_rclcpp/test_executor__rmw_cyclonedds_cpp/spin_some_max_duration/.

Considering that the Windows scheduler time slice is [20ms](https://docs.microsoft.com/en-us/windows/win32/procthread/multitasking), the test is being too optimistic (same applies for macOS and Linux, though it was only failing on Windows).

I increased all durations, to ensure that the OS time slice doesn't represent a big fraction of it.

### DIFF
--- a/test_rclcpp/test/test_executor.cpp
+++ b/test_rclcpp/test/test_executor.cpp
@@ -77,7 +77,7 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), spin_some_max_duration) {
   executor.spin_some(max_duration);
   const auto end = std::chrono::steady_clock::now();
   ASSERT_LT(max_duration, end - start);
-  ASSERT_GT(max_duration + 200ms, end - start);
+  ASSERT_GT(max_duration + 500ms, end - start);
 }
 
 TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), multithreaded_spin_call) {
--- a/test_rclcpp/test/test_executor.cpp
+++ b/test_rclcpp/test/test_executor.cpp
@@ -64,8 +64,8 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), spin_some_max_duration) {
       std::this_thread::sleep_for(100ms);
     };
   std::vector<std::shared_ptr<rclcpp::WallTimer<decltype(lambda)>>> timers;
-  // creating 20 timers which will try to do 1 ms of work each
-  // only about 10ms worth of them should actually be performed
+  // creating 20 timers which will try to do 100 ms of work each
+  // only about 1s worth of them should actually be performed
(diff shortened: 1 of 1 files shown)


test_rclcpp/test/test_executor.cpp
```diff
@@ -61,7 +61,7 @@ TEST(CLASSNAME(test_executor, RMW_IMPLEMENTATION), spin_some_max_duration) {
   rclcpp::executors::SingleThreadedExecutor executor;
   auto node = rclcpp::Node::make_shared("spin_some_max_duration");
   auto lambda = []() {
-      std::this_thread::sleep_for(1ms);
+      std::this_thread::sleep_for(100ms);
```
clalancette, line 64: This is a good change, and keeps the current semantics of the test; basically, that 20 threads of N ms sleep take < 20*N ms to complete.  However, those semantics are a bit hard to discern here.  Would you mind adding a comment explaining that?

ivanpauno: There was already a comment, but I forgot to update it.
See https://github.com/ros2/system_tests/pull/430/commits/2d0e91cc77c569049844ad410df4a6fa3956b728.

ivanpauno:
CI, testing only `test_rclcpp` `test_executor` test.
Using `--retest-until-fail 10` to check if this actually makes the test more reliable.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10612/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6052/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8642/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10507/)

ivanpauno:
The test seemed to be "less" flaky in the above run, but it still failed after a bunch of repetitions.
I incremented the time tolerance, to see if it doesn't fail anymore:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10537/)

Based on internet speculations (as windows doesn't officially clarify anything), it seems that windows server distributions have a bigger time slice that desktop versions. That's (maybe) why time based tests started failing more frequently when we switched to containerized builds.
In windows performance options, there's even a `Processor scheduling` option, where you can choose between "adjust performance for Programs"/"adjust performance for Backgroud services".
The default in Windows Server installations is the latter, though in desktops is the former.

ivanpauno:
@clalancette let me know if you think this is ready.
I can further increase time tolerance to 500ms, as timing isn't really the point of the test.

ivanpauno:
CI:
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10540/)

PR #433, opened 2020-05-27
Disable Connext security tests on Windows

jacobperron:
Connext needs a different version of OpenSSL (1.0.2n) than the system
version.
Disabling these tests until we can figure out how to run them on CI. For more context see 
 https://github.com/ros2/ci/pull/421/ and https://github.com/ros2/ci/pull/454

CI with changes in https://github.com/ros2/ci/pull/454: 
* Windows: [Build Status](https://ci.ros2.org/job/ci_windows/10806/)
* Windows Debug: [Build Status](https://ci.ros2.org/job/ci_windows/10808/)

PR #439, opened 2020-06-30
Use RTI_OPENSSL only when needed

mikaelarguedas:
The addition of openssl 1.0.2 to the libs and path was done unconditionally and the changes happening in one macro were leaking to the next macro invocation. This makes all the assignments in a single place, we could also change some macros to functions to avoid local changes to CMake variables to outlive the invocation of the macro.


mikaelarguedas:
@nuclearsandwich I removed the commit reverting #433 so this PR now is just moving the existing logic around OpenSSL 1.0.2 libraries to use them only with Connext and not leak to other test configurations (that use the system one OpenSSL 1.1.1*)

mikaelarguedas:
@nuclearsandwich friendly :bellhop_bell:

PR #454, opened 2020-12-01
Call rclcpp::init and rclcpp::shutdown in each test for test_rclcpp

brawner:
We're seeing many hangs on windows CI jobs in these tests, and I think the issue can be resolved by ensuring that `rclcpp::shutdown` gets called for each invocation of `rclcpp::init`. In the process of addressing this, I noticed that the test_remote_parameters wasn't setting up the test cases correctly for each rmw_implementation.

There may be some uncrustify issues to cleanup.

Testing `--packages-select test_rclcpp`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13192/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8123/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10908/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13206/)

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/test_rclcpp/test/test_spin.cpp
+++ b/test_rclcpp/test/test_spin.cpp
@@ -33,11 +33,23 @@
 
 using namespace std::chrono_literals;
 
+class CLASSNAME (test_spin, RMW_IMPLEMENTATION) : public ::testing::Test {
+public:
+  void SetUp() {
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDownTestCase() {
+    if (rclcpp::ok()) {
+      rclcpp::shutdown();
+    }
+  }
+};
+
 /*
    Ensures that the timeout behavior of spin_until_future_complete is correct.
  */
-TEST(CLASSNAME(test_spin, RMW_IMPLEMENTATION), test_spin_until_future_complete_timeout) {
-  if (!rclcpp::ok()) {rclcpp::init(0, nullptr);}
+TEST_F(CLASSNAME(test_spin, RMW_IMPLEMENTATION), test_spin_until_future_complete_timeout) {
   using rclcpp::FutureReturnCode;
   rclcpp::executors::SingleThreadedExecutor executor;
 
@@ -124,8 +136,7 @@ TEST(CLASSNAME(test_spin, RMW_IMPLEMENTATION), test_spin_until_future_complete_t
   }
 }
 
-TEST(CLASSNAME(test_spin, RMW_IMPLEMENTATION), spin_until_future_complete) {
-  if (!rclcpp::ok()) {rclcpp::init(0, nullptr);}
+TEST_F(CLASSNAME(test_spin, RMW_IMPLEMENTATION), spin_until_future_complete) {
   auto node = rclcpp::Node::make_shared("test_spin");
 
   // Construct a fake future to wait on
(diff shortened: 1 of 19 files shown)


test_rclcpp/test/test_spin.cpp
```diff
@@ -33,11 +33,23 @@
...
 using namespace std::chrono_literals;
 
+class CLASSNAME (test_spin, RMW_IMPLEMENTATION) : public ::testing::Test {
+public:
+  void SetUp() {
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDownTestCase() {
+    if (rclcpp::ok()) {
```
clalancette, line 43: For most of the other implementations, you just unconditionally call `rclcpp::shutdown`.  I guess this is a little better in that we won't try to `shutdown` if `init` failed somehow.  But I think that is pretty unlikely, so I'd just drop the `rclcpp::ok` check here.  Either way, I think we should be consistent among these tests.

test_rclcpp/test/test_spin.cpp
```diff
@@ -33,11 +33,23 @@
 
 using namespace std::chrono_literals;
 
+class CLASSNAME (test_spin, RMW_IMPLEMENTATION) : public ::testing::Test {
+public:
+  void SetUp() {
```
clalancette, line 38: Should this be:

```
static void SetUpTestCase() {
```

To be like the rest of the tests?

test_rclcpp/test/test_spin.cpp
```diff
@@ -33,11 +33,23 @@
...
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDownTestCase() {
```
clalancette, line 42: Should this be:

```
static void TearDownTestCase() {
```

To be like the rest of the tests?

test_rclcpp/test/test_timer.cpp
```diff
@@ -26,8 +26,20 @@
...
-TEST(CLASSNAME(test_time, RMW_IMPLEMENTATION), timer_fire_regularly) {
-  if (!rclcpp::ok()) {rclcpp::init(0, nullptr);}
+class CLASSNAME (test_time, RMW_IMPLEMENTATION) : public ::testing::Test {
+public:
+  void SetUp() {
+    rclcpp::init(0, nullptr);
+  }
+
+  void TearDown() {
+    if (rclcpp::ok()) {
```
clalancette, line 36: Same question as above; do we really need to check `rclcpp::ok` here?

clalancette (commented):
I've got some questions inline.  Also, the Rpr job is unhappy for some reason.

test_rclcpp/test/test_spin.cpp
brawner: In this case, the test `spin_until_future_complete_interrupted` calls shutdown to interrupt an executor. So shutdown may or may not have been called. I think for most other files it should remain without the check so that we'll notice if there is an issue somewhere.

test_rclcpp/test/test_timer.cpp
brawner: In this test, `timer_during_wait` calls shutdown to stop a `spin()`, but `executor.cancel()` can be used instead. Switched.

brawner (commented):
Sorry about the messy second commit. I didn't fully vet the first commit before pushing because it was taking eons to run and test on windows.

brawner:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/13203/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8134/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10919/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/13221/)

brawner:
I discovered that the `multi_consumer_clients` test in `test_multithreaded` fails for cylonedds on my machine, because it exceeded cylonedds's hardcoded core count. I added a comment to the test describing the finding and a link to the relevant issue.

https://github.com/ros2/rmw_cyclonedds/issues/268

Windows is unstable because of one cmake warning in fastcdr, but the tests otherwise pass.

Blast545:
@Mergifyio backport foxy

PR #462, opened 2021-02-19
Adapt tests to statically typed parameters

ivanpauno:
Will be needed after https://github.com/ros2/rclcpp/pull/1522 to avoid deprecation warnings.

### DIFF
--- a/test_rclcpp/test/parameter_fixtures.hpp
+++ b/test_rclcpp/test/parameter_fixtures.hpp
@@ -51,7 +51,7 @@ void declare_test_parameters(std::shared_ptr<rclcpp::Node> node, int declare_up_
   }
 
   for (int i = 0u; i < declare_up_to; ++i) {
-    node->declare_parameter(parameters[i].get_name());
+    node->declare_parameter(parameters[i].get_name(), parameters[i].get_parameter_value());
   }
 }
 
--- a/test_rclcpp/test/test_local_parameters.cpp
+++ b/test_rclcpp/test/test_local_parameters.cpp
@@ -153,7 +153,6 @@ public:
         // Check to see if they were set.
         size_t i = 0;
         for (auto & result : future.get()) {
-          fprintf(stderr, "param: %zu\n", i++);
           ASSERT_TRUE(result.successful);
         }
         executor.cancel();
--- a/test_rclcpp/test/test_local_parameters.cpp
+++ b/test_rclcpp/test/test_local_parameters.cpp
@@ -151,7 +151,6 @@ public:
       {
         printf("Got set_parameters result\n");
         // Check to see if they were set.
-        size_t i = 0;
         for (auto & result : future.get()) {
           ASSERT_TRUE(result.successful);
         }
--- a/test_rclcpp/test/test_local_parameters.cpp
+++ b/test_rclcpp/test/test_local_parameters.cpp
@@ -121,12 +121,12 @@ public:
   ParametersAsyncNode()
   : Node("test_local_parameters_async_with_callback")
   {
-    this->declare_parameter("foo");
-    this->declare_parameter("bar");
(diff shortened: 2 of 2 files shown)


test_rclcpp/test/parameter_fixtures.hpp
```diff
@@ -51,7 +51,7 @@ void declare_test_parameters(std::shared_ptr<rclcpp::Node> node, int declare_up_
   }
 
   for (int i = 0u; i < declare_up_to; ++i) {
-    node->declare_parameter(parameters[i].get_name());
+    node->declare_parameter(parameters[i].get_name(), parameters[i].get_parameter_value());
```
jacobperron, line 54: What about declaring with the type instead?

```suggestion
    node->declare_parameter(parameters[i].get_name(), parameters[i].get_type());
```

Then I think there'd be less chance of interfering with the tests if they are also checking the value.

test_rclcpp/test/test_local_parameters.cpp
```diff
@@ -151,7 +151,9 @@ class ParametersAsyncNode : public rclcpp::Node
       {
         printf("Got set_parameters result\n");
         // Check to see if they were set.
+        size_t i = 0;
         for (auto & result : future.get()) {
+          fprintf(stderr, "param: %zu\n", i++);
```
jacobperron, line 156: This looks vestigial.

test_rclcpp/test/parameter_fixtures.hpp
ivanpauno: When declaring only with the type, an user provided override is needed.
Between a default and an override there is no difference.

The other option is to modify the descriptor to make the parameter dynamically typed.

test_rclcpp/test/test_local_parameters.cpp
ivanpauno:
```suggestion
```

test_rclcpp/test/test_local_parameters.cpp
```diff
@@ -151,6 +151,7 @@ class ParametersAsyncNode : public rclcpp::Node
       {
         printf("Got set_parameters result\n");
         // Check to see if they were set.
+        size_t i = 0;
```
jacobperron, line 154: This statement can be removed  too,

test_rclcpp/test/parameter_fixtures.hpp
jacobperron: oh, I thought that an override was only needed if the code tries to get the parameter before it is set. Is that not the case?

test_rclcpp/test/test_local_parameters.cpp
ivanpauno:
```suggestion
```

test_rclcpp/test/parameter_fixtures.hpp
ivanpauno: I think we settled on throwing an exception in the declaration in this discussion https://github.com/ros2/rcl_interfaces/pull/118#discussion_r563976030

jacobperron: You're right, I forgot about that compromise :grin: 

I think it's worth documenting the design decisions (e.g. why we went this this API), based on our prior discussions. Then we'll have something to reference when similar questions about this come up in the future. Perhaps as a quick note in a new `doc` directory in rcl_interfaces repo, or added to this design page: http://design.ros2.org/articles/ros_parameters.html

jacobperron (approved):

test_rclcpp/test/parameter_fixtures.hpp
ivanpauno: Sounds good, I will open a new PR with docs

PR #465, opened 2021-03-10
Make test pass after rclcpp#1532

ivanpauno:
Some of the tests were susceptible at the amount of parameters declared, and https://github.com/ros2/rclcpp/pull/1532 is adding new parameters. That was fixed in https://github.com/ros2/system_tests/commit/f44712759998ca7c257b110f0fd4deac28e87724.

https://github.com/ros2/system_tests/commit/11d32755043faf733ce4e0f6993d46b69ad9bcf0 is a hack to fix a flaky test, I don't have a better way to test that in mind.

### DIFF
--- a/test_rclcpp/test/test_timeout_subscriber.cpp
+++ b/test_rclcpp/test/test_timeout_subscriber.cpp
@@ -56,6 +56,9 @@ TEST_F(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscribe
     "test_message_timeout_uint32", 10, callback);
 
   rclcpp::executors::SingleThreadedExecutor executor;
+  executor.add_node(node);
+
+  executor.spin_all(std::chrono::milliseconds{100});  // clear any previous ready executable
 
   size_t num_cycles = 5;
 
@@ -64,7 +67,7 @@ TEST_F(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscribe
 
     // ensure that the non-blocking spin does return immediately
     auto nonblocking_start = std::chrono::steady_clock::now();
-    executor.spin_node_once(node, std::chrono::milliseconds::zero());
+    executor.spin_once(std::chrono::milliseconds::zero());
     auto nonblocking_end = std::chrono::steady_clock::now();
     auto nonblocking_diff = nonblocking_end - nonblocking_start;
     EXPECT_LT(
@@ -74,7 +77,7 @@ TEST_F(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscribe
     // ensure that the blocking spin does return after the specified timeout
     auto blocking_timeout = std::chrono::milliseconds(100);
     auto blocking_start = std::chrono::steady_clock::now();
-    executor.spin_node_once(node, blocking_timeout);
+    executor.spin_once(blocking_timeout);
     auto blocking_end = std::chrono::steady_clock::now();
     auto blocking_diff = blocking_end - blocking_start;
     EXPECT_GT(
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/test_timeout_subscriber.cpp
```diff
@@ -74,7 +77,7 @@ TEST_F(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscribe
     // ensure that the blocking spin does return after the specified timeout
     auto blocking_timeout = std::chrono::milliseconds(100);
     auto blocking_start = std::chrono::steady_clock::now();
-    executor.spin_node_once(node, blocking_timeout);
+    executor.spin_once(blocking_timeout);
```
wjwwood, line 80: For this test, I think you'd have to:

- time a spin once
- if that spin once called the callback, check how long
- if not, loop

Basically spin one thing at a time, recording the time it took, and then check to see if it was the spin once we were looking for.

wjwwood (commented):
The changes to the parameter tests look ok.

test_rclcpp/test/test_timeout_subscriber.cpp
ivanpauno: mmm, I don't think this was testing that.
The callback is only throwing an exception https://github.com/ros2/system_tests/blob/11d32755043faf733ce4e0f6993d46b69ad9bcf0/test_rclcpp/test/test_timeout_subscriber.cpp#L33

IIUC what this is trying to test is that the timeout argument of `spin_once` is being used correctly.
But for actually checking that you have to make sure that no executable will be ready, which is pretty hard.

Honestly, I don't know why https://github.com/ros2/rclcpp/pull/1532 made this fail, seems to be random.

---

What I could do is spinning without adding any node to the executor ... I will try that.

ivanpauno: > What I could do is spinning without adding any node to the executor ... I will try that.

That also works, but really IDK if that was the intention of this test.

wjwwood: To be honest, I think the original goal of the test is a bit misguided.

Maybe we could create a callback group for the subscription, add that the executor, then do the test as normal, therefore the things in the node will not affect it?

ivanpauno: That makes sense to me, see https://github.com/ros2/system_tests/pull/465/commits/f71c5524be5d034991b413dc5e4967badc878cef.

test_rclcpp/test/test_timeout_subscriber.cpp
```diff
@@ -51,11 +51,15 @@ TEST_F(CLASSNAME(test_timeout_subscriber, RMW_IMPLEMENTATION), timeout_subscribe
   auto start = std::chrono::steady_clock::now();
 
   auto node = rclcpp::Node::make_shared("test_timeout_subscriber");
+  auto cg = node->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
```
jacobperron, line 54: It might be worth leaving a comment here for posterity. Otherwise, someone might come back and refactor this to the way it was before.

jacobperron (approved):
LGTM

test_rclcpp/test/test_timeout_subscriber.cpp
ivanpauno: what would be the comment?
Really I still don't get the goal of the test :joy:, this change just makes it not fail randomly.

jacobperron: Maybe `// Add subscription to its own callback group to avoid interference from other things in the node` ?

ivanpauno:
```suggestion
  // Add subscription to its own callback group to avoid interference from other things in the node
  auto cg = node->create_callback_group(rclcpp::CallbackGroupType::MutuallyExclusive);
```

ivanpauno: done in https://github.com/ros2/system_tests/pull/465/commits/f095533e21eaf4c7f81d1e9ee9d1e998ad1abef1

wjwwood (approved):

ivanpauno:
CI is [here](https://github.com/ros2/rclcpp/pull/1532#issuecomment-797481908), going in!

PR #478, opened 2021-07-13
Add test for defered service callback signature

ivanpauno:
See https://github.com/ros2/rclcpp/pull/1709

### DIFF
--- a/test_rclcpp/test/test_services_server.cpp
+++ b/test_rclcpp/test/test_services_server.cpp
@@ -52,9 +52,10 @@ public:
         if (handle_defered_response_thread_.joinable()) {
           throw std::runtime_error{"expected the callback to be called only once"};
         }
+        auto shared_this = this->shared_from_this();
         handle_defered_response_thread_ = std::thread(
           [
-            me = this->shared_from_this(),
+            me = std::move(shared_this),
             request = std::move(request),
             request_header = std::move(request_header)]()
           {
--- a/test_rclcpp/test/test_services_server.cpp
+++ b/test_rclcpp/test/test_services_server.cpp
@@ -95,15 +95,16 @@ int main(int argc, char ** argv)
   auto defered_cb = std::make_shared<DeferedCbServiceWrapper>();
   defered_cb->create_service(*node);
 
+  rclcpp::TimerBase::SharedPtr timer;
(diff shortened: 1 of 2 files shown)


test_rclcpp/test/test_services_server.cpp
```diff
@@ -34,6 +34,27 @@ void handle_add_two_ints_reqid(
...
+  {
+    impl_ = node.create_service<test_rclcpp::srv::AddTwoInts>(
+      "add_two_ints_defered_cb",
+      [this](
+        const std::shared_ptr<rmw_request_id_t> request_header,
+        const std::shared_ptr<test_rclcpp::srv::AddTwoInts::Request> request)
+      {
+        test_rclcpp::srv::AddTwoInts::Response response;
+        response.sum = request->a + request->b;
+        this->impl_->send_response(*request_header, response);
```
wjwwood, line 50: Should we do this from a thread or something? I'm just wondering if sending the response before the callback is done is some how hiding potential bugs. Also, it seems like users would want to send the response asynchronously after the callback, otherwise they'd just return the response (or set it from the arguments).

ivanpauno: > Should we do this from a thread or something? I'm just wondering if sending the response before the callback is done is some how hiding potential bugs. Also, it seems like users would want to send the response asynchronously after the callback, otherwise they'd just return the response (or set it from the arguments).

I agree that users are probably not going to use this this way, but I didn't want to make the test to complex.
Adding a thread sounds reasonable though

ivanpauno: See https://github.com/ros2/system_tests/pull/478/commits/81703c250838f76bb52c8aa463cb8eaa1297b669.

wjwwood (approved):

ivanpauno:
CI: https://github.com/ros2/rclcpp/pull/1709#issuecomment-881736570

wjwwood (approved):

ivanpauno:
Last commit should've fixed the CI issues

ivanpauno:
Finally got windows CI passing: https://github.com/ros2/rclcpp/pull/1709#issuecomment-883667541.

wjwwood (approved):

ivanpauno:
@ros-pull-request-builder retest this please

PR #489, opened 2021-11-08
Updated maintainers

adityapande-1995:
Signed-off-by: Aditya Pande <aditya050995@gmail.com>

### DIFF
--- a/test_cli/package.xml
+++ b/test_cli/package.xml
@@ -6,11 +6,13 @@
   <description>
     Test command line arguments passed to ros2 executables.
   </description>
-  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
-  <maintainer email="william@openrobotics.org">William Woodall</maintainer>
+  <maintainer email="aditya.pande@osrfoundation.org">Aditya Pande</maintainer>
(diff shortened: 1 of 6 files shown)


test_cli/package.xml
```diff
@@ -6,11 +6,13 @@
   <description>
     Test command line arguments passed to ros2 executables.
   </description>
-  <maintainer email="jacob@openrobotics.org">Jacob Perron</maintainer>
-  <maintainer email="william@openrobotics.org">William Woodall</maintainer>
+  <maintainer email="aditya.pande@osrfoundation.org">Aditya Pande</maintainer>
+  <maintainer email="shane@osrfoundation.org">Shane Loretz</maintainer>
```
jacobperron, line 10: I believe we have a preference to use the "openrobotics.org" domain ([for example](https://github.com/ros2/rclpy/blob/f2cb25b215781bd673457dd4feb82c98b7ffcb70/rclpy/package.xml#L8)).

adityapande-1995: Done !

jacobperron (approved):
I we can skip running CI from ci.ros2.org for this PR. The Rpr job passing is enough.

clalancette (approved):

PR #515, opened 2023-04-14
refactor the multi_access_publisher test to avoid dead locks

wjwwood:
This is based on discussions with @clalancette and @mjcarroll about this test failing some times on low-core machines, the theory being that the timers all are waiting on the subscriptions, but doing so by blocking and therefore using all the executor threads. This change mainly changes this by having the timers give up after publishing everything and then having a separate thread to monitor the test and cancel the executor when done.

### DIFF
--- a/test_rclcpp/test/test_multithreaded.cpp
+++ b/test_rclcpp/test/test_multithreaded.cpp
@@ -324,9 +324,9 @@ static inline void multi_access_publisher(bool intra_process)
   std::atomic_uint subscription_counter(0);
   auto sub_callback = [&subscription_counter](const test_rclcpp::msg::UInt32::ConstSharedPtr msg)
     {
-      ++subscription_counter;
-      printf("Subscription callback %u\n", subscription_counter.load());
-      printf("callback() %d with message data %u\n", subscription_counter.load(), msg->data);
+      auto this_subscription_count = subscription_counter++;
+      printf("Subscription callback %u\n", this_subscription_count);
+      printf("callback() %d with message data %u\n", this_subscription_count, msg->data);
     };
   auto sub_callback_group = node->create_callback_group(
     rclcpp::CallbackGroupType::Reentrant);
@@ -342,22 +342,22 @@ static inline void multi_access_publisher(bool intra_process)
   test_rclcpp::wait_for_subscriber(node, node_topic_name);
 
   // timers
-  std::atomic_uint timer_counter(0);
+  std::atomic_uint publish_counter(0);
   auto timer_callback =
-    [&executor, &pub, &timer_counter, &subscription_counter, &num_messages](
+    [&executor, &pub, &publish_counter, &subscription_counter, &num_messages](
     rclcpp::TimerBase & timer)
     {
       auto msg = std::make_unique<test_rclcpp::msg::UInt32>();
-      auto next_timer_count = timer_counter.fetch_add(1);
-      if (next_timer_count >= num_messages) {
-        // enough messages have been sent, step counter back and cancel timer
-        timer_counter--;
+      auto next_publish_count = ++publish_counter;
+      if (next_publish_count > num_messages) {
+        // enough messages have been sent, step the counter back, and cancel timer
+        publish_counter--;
         timer.cancel();
         return;
       }
       // publish a new message
-      msg->data = next_timer_count;
-      printf("Publishing message %u\n", timer_counter.load());
+      msg->data = next_publish_count;
+      printf("Publishing message %u\n", msg->data);
       pub->publish(std::move(msg));
     };
   std::vector<rclcpp::TimerBase::SharedPtr> timers;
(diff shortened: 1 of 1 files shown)


wjwwood:
CI just up to `test_rclcpp` and only testing `test_rclcpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18568/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13113/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19306/)

clalancette:
Given that we are worried about this test being flaky, can I suggest running CI on this over the weekend with something like `--retest-until-fail 100` or something?  Also on Windows Debug and RHEL.

wjwwood:
Ok, I'll do those if there's no comments on review.

wjwwood:
I fixed some more corner cases in [04eb559](https://github.com/ros2/system_tests/pull/515/commits/04eb55970ba0ccf82172808ac6d3ef852dd38e4a), the retest until fail was a good call.

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -308,40 +338,72 @@ static inline void multi_access_publisher(bool intra_process)
...
-    [&executor, &pub, &timer_counter, &subscription_counter, &num_messages](
+    [&executor, &pub, &publish_counter, &subscription_counter, &num_messages](
     rclcpp::TimerBase & timer)
     {
       auto msg = std::make_unique<test_rclcpp::msg::UInt32>();
-      if (timer_counter.load() >= num_messages) {
+      auto next_publish_count = ++publish_counter;
+      if (next_publish_count > num_messages) {
+        // enough messages have been sent, step the counter back, and cancel timer
+        publish_counter--;
```
clalancette, line 354: Hm, this bit feels non-atomic to me.  I understand that this should only happen after we've published enough messages, and I honestly can't come up with a scenario where this fails.  But do we really need to do this decrement?  Could we instead just let it increment too high unnecessarily, then use `num_messages` when doing checks below?

clalancette (commented):
I have one comment/question, this otherwise looks good to me.

test_rclcpp/test/test_multithreaded.cpp
clalancette: > Could we instead just let it increment too high unnecessarily, then use num_messages when doing checks below?

Actually, sorry, this is dumb.  This won't work because we are comparing `num_messages` to `publish_counter`.

That said, I wonder if we can just drop the `publish_counter` checks below, and instead just rely on the `subscription_counter`.  At the end of the day, that's the thing we are most interested in.  Thoughts?

wjwwood: Well, I could do that, though it's nice to see that we actually did publish the right number of times, and that accessing the counter for that publish is thread-safe (needing the atomics and this back and forth business). We could use a lock to check the value before incrementing it (avoiding the roll back) but I was trying to keep with the atomics.

I'd like to merge this incremental improvement and open a new pr for using a lock (should be clearer to read), but I want this to go in before the freeze ideally.

wjwwood:
Linux retest until fail 100: [Build Status](https://ci.ros2.org/job/ci_linux/18589/)

mjcarroll:
There are failures here, none of which are the one you were trying to fix :)

wjwwood:
Though the 100 retest job is yellow, it's due to other tests in `test_rclcpp` that are presumably flaky.

test_rclcpp/test/test_multithreaded.cpp
wjwwood: Ok, I think I resolved your concern in [b7c988e](https://github.com/ros2/system_tests/pull/515/commits/b7c988ee8ad7a057a52c9c7d9bf2938a8812327d), at least it's clearer to me. Though I'm still checking all three "counters".

wjwwood:
CI with build args `--packages-up-to test_rclcpp` and test args `--packages-select test_rclcpp`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18601/)
* Linux (`--retest-until-fail 100`) [Build Status](http://ci.ros2.org/job/ci_linux/18602/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13139/)
* RHEL [Build Status](https://ci.ros2.org/job/ci_linux-rhel/407/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19336/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/19337/)

I also tested locally against ros2/rclcpp@rolling and ros2/rclcpp@mjcarroll/rclcpp_waitset_executor for the pr https://github.com/ros2/rclcpp/pull/2142, both are passing for me.

mjcarroll:
LGTM, but you got busted by Windows warnings

wjwwood:
I have never seen something compile on Linux but not Linux-aarch64 🤯 

I'll have to look into that.

clalancette:
So at the end of the day, the problem was that `std::numeric_limits::max` needs to be a function call, not a member variable.  Fixing that fixed it for me.  While I was in there, I also did a couple of minor updates to switch types to a uint32_t so we can get rid of a cast.

That said, I have no idea why this worked at all on Linux amd64 and Windows.  It should have failed there as well.  In any case, I'll run another CI on it now to see what it looks like.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18649/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13178/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19378/)

mjcarroll (approved):

test_rclcpp/test/test_multithreaded.cpp
```diff
@@ -308,40 +349,80 @@ static inline void multi_access_publisher(bool intra_process)
...
+          return;
         }
-        executor.cancel();
-        return;
+        next_publish_count = publish_counter++;
       }
-      msg->data = ++timer_counter;
-      printf("Publishing message %u\n", timer_counter.load());
+      // publish a new message
+      ASSERT_LE(next_publish_count, std::numeric_limits<uint32_t>::max());
```
wjwwood, line 373: This assert was only needed when doing the static cast from `size_t`.

PR #514, opened 2023-04-06
Fix ODR errors with gtest

methylDragon:
`test_quality_of_service` is explicitly linking against gtest, but it gets linked again as a result of the `ament_add_gtest` call.

This causes an ODR violation, but it didn't seem to cause any issues with tests.
Running the QOS tests with asan causes ODR violations though.

PR #518, opened 2023-06-25
Fix test QoS on macOS by re-adding target_link_libraries *_support GTEST_LIBRARIES

### LINKED ISSUE #517: [Test QoS] macOS needs target_link_libraries GTEST_LIBRARIES
## Bug report

Compiliing [ROS2 Iron from source on macOS](https://docs.ros.org/en/iron/Installation/Alternatives/macOS-Development-Setup.html) gets an error after fded516933cca6f5bafc64bac489c86dfee4e323.

**Required Info:**

- Operating System:
  - macOS 13.4 Ventura (arm64)
- Installation type:
  - from source
- Version or commit hash:
  - iron @ 6324429e11b13287e77269c6ab39d17509ec77ce
- DDS implementation:
  - default: Fast DDS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

```console
# follow ROS2 iron compile from source instructions

mkdir -p ~/ros2_iron/src
cd ~/ros2_iron
vcs import --input https://raw.githubusercontent.com/ros2/ros2/iron/ros2.repos src

cd ~/ros2_iron/
colcon build --symlink-install --packages-skip-by-dep python_qt_binding
```

#### Expected behavior

* system_tests compiles like other repositories.

#### Actual behavior

```console
--- stderr: test_quality_of_service                                                                                                                                                                                          
Undefined symbols for architecture arm64:
  "typeinfo for testing::Test", referenced from:
      typeinfo for BaseQosRclcppTestFixture in qos_utilities.cpp.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libtest_quality_of_service_support.dylib] Error 1
make[1]: *** [CMakeFiles/test_quality_of_service_support.dir/all] Error 2
make: *** [all] Error 2
make: INTERNAL: Exiting with 11 jobserver tokens available; should be 10!
---
Failed   <<< test_quality_of_service [11.5s, exited with code 2]
```

#### Additional information

Re-inserting `target_link_libraries(${PROJECT_NAME}_support ${GTEST_LIBRARIES})` (test_quality_of_service/CMakeLists.txt:47), makes it compile. Maybe we should re-add it or re-add it using `if (APPLE)`.

PhDittmann:
Fixes #517 by re-adding `target_link_libraries(${PROJECT_NAME}_support ${GTEST_LIBRARIES})` to `test_quality_of_service/CMakeLists.txt:47`.

Tested with macOS Ventura 13.4 arm64 and Docker container [arm64v8/ubuntu:latest](https://hub.docker.com/r/arm64v8/ubuntu/)

### DIFF
--- a/test_quality_of_service/CMakeLists.txt
+++ b/test_quality_of_service/CMakeLists.txt
@@ -39,11 +39,9 @@ if(BUILD_TESTING)
     test/qos_test_node.cpp
     test/qos_test_publisher.cpp
     test/qos_test_subscriber.cpp
-    test/qos_utilities.cpp
   )
 
   ament_find_gtest()
-  target_include_directories(${PROJECT_NAME}_support PUBLIC "${GTEST_INCLUDE_DIRS}")
   target_compile_definitions(${PROJECT_NAME}_support PRIVATE
     "TEST_QUALITY_OF_SERVICE_BUILDING_LIBRARY")
   ament_target_dependencies(${PROJECT_NAME}_support rclcpp rcutils std_msgs)
@@ -61,10 +59,10 @@ if(BUILD_TESTING)
   endfunction()
 
   macro(tests)
-    add_custom_gtest(test_deadline test/test_deadline.cpp)
-    add_custom_gtest(test_lifespan test/test_lifespan.cpp)
-    add_custom_gtest(test_liveliness test/test_liveliness.cpp)
-    add_custom_gtest(test_best_available test/test_best_available.cpp)
+    add_custom_gtest(test_deadline test/test_deadline.cpp test/qos_utilities.cpp)
+    add_custom_gtest(test_lifespan test/test_lifespan.cpp test/qos_utilities.cpp)
+    add_custom_gtest(test_liveliness test/test_liveliness.cpp test/qos_utilities.cpp)
+    add_custom_gtest(test_best_available test/test_best_available.cpp test/qos_utilities.cpp)
   endmacro()
 
   call_for_each_rmw_implementation(tests)
--- a/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
+++ b/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp
@@ -40,7 +40,6 @@
  * @param milliseconds
  * @return tuple of milliseconds converted to <seconds, nanoseconds>
  */
-TEST_QUALITY_OF_SERVICE_PUBLIC
 std::tuple<size_t, size_t> convert_chrono_milliseconds_to_size_t(
   const std::chrono::milliseconds & milliseconds);
 
@@ -67,9 +66,7 @@ bool wait_for(
 class BaseQosRclcppTestFixture : public ::testing::Test
 {
 protected:
-  TEST_QUALITY_OF_SERVICE_PUBLIC
   void SetUp() override;
-  TEST_QUALITY_OF_SERVICE_PUBLIC
   void TearDown() override;
   // executor used to submit work
   std::shared_ptr<rclcpp::executors::SingleThreadedExecutor> executor;


test_quality_of_service/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ if(BUILD_TESTING)
 
   ament_find_gtest()
   target_include_directories(${PROJECT_NAME}_support PUBLIC "${GTEST_INCLUDE_DIRS}")
+  target_link_libraries(${PROJECT_NAME}_support ${GTEST_LIBRARIES})
```
clalancette, line 47: So we ended up removing this in #514 because we were having ODR issues on Linux and Windows.  We'll need to make sure that that doesn't happen again.

But that said, I think a better way to go here is to actually get rid of the need for `gtest` in this support library at all.  Only `BaseQosRclcppTestFixture` in `qos_utilities.{cpp,hpp}` ends up using it.  So I think what we should probably do instead is to remove that particular class from that utility method, and just replicate it into the 4 places that need it (`test_deadline.cpp`, `test_lifespan.cpp`, `test_liveliness.cpp`, and `test_best_available.cpp`).  Then we can also remove the `target_include_directories` above, and that solution should definitely avoid ODR and also make this work for macOS.

That said, I actually don't see why we need this at all.  As far as I can tell, none of the source files that make up `${PROJECT_NAME}_support` actually depend on `gtest`, so they shouldn't link against it.  In point of fact, I would also remove the `target_include_directories` line above, since that shouldn't be needed either.

clalancette (changes requested):
Thanks for looking into this!  I've left a detailed comment inline on what I think we should do instead.

test_quality_of_service/CMakeLists.txt
PhDittmann: You're right: qos_test_{node|publisher|subscriber} don't need it at all. Instead of copying BaseQosRclcppTestFixture to the four tests, in CMakeLists.txt I moved `qos_utilities.cpp` from `${PROJECT_NAME}_support` to the `add_custom_gtest` calls. This avoids the suggested code redundancy, but it needs to be compiled 4 times though.

clalancette: Yeah, good call.  This will do.

Can you please add a Signed-off-by line to your commit (`git commit --amend -s`)?  Once that is done, I'll approve and run CI on it.

PhDittmann: Of course. It's added and pushed.

PhDittmann: The ubuntu jammy amd64 test fails because of latest commit https://github.com/ros2/system_tests/commit/d58707976ef51d049288b9d5bf9da681fe51ced1. Shall I rebase to previous commit or is it fine?

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19045/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13562/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19789/)

clalancette (changes requested):
The failing CI on Linux was just due to infrastructure issues.  I'll restart that one again later.

The Windows build is a little unhappy, though.  In particular, it is saying "inconsistent dll linkage".  I think the solution here is that since we are no longer compiling this as a library, we need to remove https://github.com/ros2/system_tests/blob/d58707976ef51d049288b9d5bf9da681fe51ced1/test_quality_of_service/include/test_quality_of_service/qos_utilities.hpp#L43 .  Once that is done, I think this should be happier and I'll run CI again.

clalancette:
Here's another try at CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19048/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13569/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19792/)

PhDittmann:
You're right. I removed the 3 occurences.

clalancette:
@mergifyio backport iron

PR #522, opened 2023-08-05
Use test fixtures to create SROS artifacts

cottsay:
The current implementation creates SROS artifacts during the configure phase. This is undesirable for a variety of reasons, such as:
* The ros2 binary is invoked during build, meaning that a regression in that tool will break the entire CI build instead of appearing as a test failure.
* Some of the environment is captured during the artifact generation and it may not be the same when the tests are invoked. Namely, changes to ROS_DOMAIN_ID between build and test will break.

The test fixtures work similarly to tests themselves, and are automatically run if any tests requiring them are enabled. If the fixture fails, the dependent tests are not attempted.

### DIFF
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -312,42 +312,51 @@ if(BUILD_TESTING)
       set(KEYSTORE_DIRECTORY_NATIVE_PATH "${KEYSTORE_DIRECTORY}")
     endif()
 
-    # generate security artifacts using sros2
-    find_program(PROGRAM ros2)
+    #
+    # CTest Fixtures
+    #
+    # These fixtures are set up as needed at the beginning of the test run to
+    # support any selected tests which require them.
+    #
+    # * sros_artifacts: This fixture generates SROS2 security artifacts needed
+    #   for the tests in this package.
+    #
 
-    set(node_names_list "/publisher;/subscriber;/publisher_missing_key;/publisher_invalid_cert")
-
-    ament_add_test(pre_clean_artifacts
+    # remove previously generated enclaves
+    add_test(NAME pre_clean_artifacts
       COMMAND ${CMAKE_COMMAND} -E rm -rf "${KEYSTORE_DIRECTORY}/enclaves/"
-      GENERATE_RESULT_FOR_RETURN_CODE_ZERO
     )
     set_tests_properties(pre_clean_artifacts PROPERTIES
       FIXTURES_SETUP sros_artifacts
     )
 
-    ament_add_test(generate_artifacts
-      COMMAND ${PROGRAM} security generate_artifacts -k "${KEYSTORE_DIRECTORY_NATIVE_PATH}" -e ${node_names_list}
-      GENERATE_RESULT_FOR_RETURN_CODE_ZERO
+    # generate security artifacts using sros2
+    find_program(PROGRAM ros2)
+
+    set(node_names_list "/publisher;/subscriber;/publisher_missing_key;/publisher_invalid_cert")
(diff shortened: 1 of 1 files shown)


cottsay:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19300/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13832/)
* Linux-rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/457/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20021/)

Prototype build with this change and parallelized tests: [Build Status](https://ci.ros2.org/job/ci_linux/19298/)

test_security/CMakeLists.txt
```diff
@@ -312,24 +316,44 @@ if(BUILD_TESTING)
...
 
     set(node_names_list "/publisher;/subscriber;/publisher_missing_key;/publisher_invalid_cert")
-    file(REMOVE_RECURSE "${KEYSTORE_DIRECTORY}/enclaves/publisher_invalid_cert")
-    set(generate_artifacts_command ${PROGRAM} security generate_artifacts -k ${KEYSTORE_DIRECTORY_NATIVE_PATH} -e ${node_names_list})
-    execute_process(
-      COMMAND ${generate_artifacts_command}
-      RESULT_VARIABLE GENERATE_ARTIFACTS_RESULT
-      ERROR_VARIABLE GENERATE_ARTIFACTS_ERROR
+
+    ament_add_test(pre_clean_artifacts
```
cottsay, line 320: These really don't need to be ament tests, but I made them such. I'm on the fence about converting them to plain ctest tests instead, especially since they can't really be 'selected' and are just run when needed.

clalancette: Yeah, from what I understand about fixtures (which isn't too much), these will always be run when they are needed, regardless of what you select.  From that perspective, I think it would make sense to make these regular ctest tests, since it is something of an indication that there is something special/different about them.

I think we could also use a comment in this section explaining why we are doing this here.

clalancette (commented):
Overall, this looks like a great cleanup.  I've left a couple of things to think about.

test_security/CMakeLists.txt
cottsay: Thanks for the review.

Alright, I converted to simple `add_test` calls and added some comments.

IMO, going into detail about why we're doing this during the test phase rather than configure phase isn't valuable. In my experience the existing approach was more of a deviation from the norm, and this approach is closer to what I've seen elsewhere in the ROS 2 codebase. That said, I did add more comments discussing what's being defined here.

cottsay:
Fresh CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19306/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13836/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20026/)

clalancette (approved):
This looks good to me with green CI.

cottsay:
Thanks for the review.

In order to really leverage this on ci.ros2.org, we'll probably want to backport it to Iron and Humble. CMake 3.7 is an acceptable limitation for both of those platforms from what I can tell, and this shouldn't have any affect on payload.

Do you see any reason not to backport?

clalancette:
> Do you see any reason not to backport?

No, I think that is totally fine.

cottsay:
@Mergifyio backport iron humble

clalancette:
@cottsay It looks like this may have lead to a regression in the nightly repeated jobs: https://ci.ros2.org/view/nightly/job/nightly_linux-aarch64_repeated/2468/testReport/(root)/projectroot/remove_missing_key/ .

Can you please take a look?

PR #525, opened 2023-08-08
Extract sros_artifacts fixture into a CMake script

cottsay:
All of these commands should be run together to set up the fixture for the tests. When running tests with repitition, we want to run through the entire artifact setup and not run each individual command multiple times.

Fixes #522

CI, using `--retest-until-fail 2`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19310/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13841/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20030/)

### DIFF
--- a/test_security/CMakeLists.txt
+++ b/test_security/CMakeLists.txt
@@ -322,44 +322,16 @@ if(BUILD_TESTING)
     #   for the tests in this package.
     #
 
-    # remove previously generated enclaves
-    add_test(NAME pre_clean_artifacts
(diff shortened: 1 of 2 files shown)


clalancette (commented):
If I understand this properly, then we are essentially treating `sros_artifacts.cmake` as a test script and executing the entirety of it before each test invocation.  Is that correct?

Could we also make this test an e.g. Python script instead?  I actually don't care that much whether we keep this as CMake or a script, but I just wanted to mention that as a possibility.

PR #528, opened 2023-09-11
Update to C++17

clalancette:


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19494/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14014/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20213/)

ahcorde (approved):

clalancette:
The test failures is a quite surprising result here.  I'll have to look into it.

mjcarroll:
Looks like many "failed to create participant", could it be a transient thing?

clalancette:
> Looks like many "failed to create participant", could it be a transient thing?

Maybe, but I also just hit that locally.

clalancette:
Here's another CI attempt (only on Linux) with some recent CI changes reverted.  This is just to check whether that makes any difference:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19545/)

PR #530, opened 2023-10-02
Add integration test for nested messages.

clalancette:
This is an integration test to ensure that https://github.com/ros2/rmw_fastrtps/issues/715 doesn't happen again.

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -26,8 +26,11 @@ ament_auto_find_build_dependencies()
 
 if(BUILD_TESTING)
   find_package(ament_cmake REQUIRED)
-  find_package(test_msgs REQUIRED)
   find_package(osrf_testing_tools_cpp REQUIRED)
+  find_package(rcpputils REQUIRED)
+  find_package(rcl REQUIRED)
+  find_package(rclcpp REQUIRED)
+  find_package(test_msgs REQUIRED)
 
   ament_index_get_resource(interface_files "rosidl_interfaces" "test_msgs")
   string(REPLACE "\n" ";" interface_files "${interface_files}")
@@ -57,6 +60,8 @@ if(BUILD_TESTING)
 
   set(other_message_files
     "msg/FieldsWithSameType.msg"
+    "msg/Inner.msg"
+    "msg/Outer.msg"
     "msg/UInt32.msg"
   )
   rosidl_generate_interfaces(${PROJECT_NAME}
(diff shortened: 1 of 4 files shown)


test_communication/CMakeLists.txt
```diff
@@ -449,6 +454,11 @@ if(BUILD_TESTING)
   custom_executable(test_action_server_cpp
     "test/test_action_server.cpp")
 
+  # Regression test for https://github.com/ros2/rmw_fastrtps/issues/715
+  custom_executable(test_nested_types
+    "test/test_nested_types.cpp")
+  target_link_libraries(test_nested_types rcpputils::rcpputils)
```
fujitatomoya, line 460: although we did not see this issue with other Tier I RMW implementations, they still need to be tested here? I think all RMW implementations need to be tested, just checking.

methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19596/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14107/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20320/)

methylDragon (approved):
👍 pending green CI

fujitatomoya (approved):

test_communication/CMakeLists.txt
methylDragon: This does test it for the other RMWs, right?
This is testing for expected, and potentially unintuitive behavior breaks anyway, so it's good to test 👍

methylDragon:
I think the gh actions CI is failing because the fix PR isn't in yet. Do we wait for the fix PR to go in before merging this?

test_communication/CMakeLists.txt
clalancette: > This does test it for the other RMWs, right?

Correct.  Because this test is listed in the `targets` macro down below, and then the `targets` macro is used in `call_for_each_rmw_implementation` even further below, it will be run with every RMW implementation.  This is done by setting the `RMW_IMPLEMENTATION` environment variable at runtime.

fujitatomoya:
@Mergifyio backport iron humble

fujitatomoya:
@clalancette since i do not have write permission on this repo, can you do me a favor?

clalancette:
> @Mergifyio backport iron humble

To be honest, I don't think we need to do this backport.  We have the test in `rolling`, which I think is good enough.  The fixes we should definitely backport.

fujitatomoya:
@clalancette sounds good to me. just thought all related PRs can be backported.

PR #532, opened 2023-11-27
Switch to target_link_libraries everywhere.

clalancette:
While we are in here, also remove some dead code relating to old RMWs that no longer exist.

A draft for now since I want to run CI to ensure that we are running all of the same tests.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19886/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14372/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20595/)

PR #535, opened 2024-02-14
Revamp test_rclcpp to compile far few files.

clalancette:
Instead of recompiling the tests multiple times for each RMW, instead just compile it once and use the RMW_IMPLEMENTATION environment variable to control which RMW gets used.  On my machine, this sped up
compilation by about 2x, and should reduce the number of times our coverage jobs fail because they run out of memory.

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -15,11 +15,18 @@ find_package(ament_cmake REQUIRED)
 
 if(BUILD_TESTING)
   find_package(rclcpp REQUIRED)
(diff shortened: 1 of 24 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20295/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14706/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21025/)

mjcarroll (approved):

PR #540, opened 2024-03-27
Rolling release.

marcoag:
Changelog update and version bump. I will folow up with the tag and bloom once this is merged.

NOTE: We might wanna rebase to keep the commits as per the history in this repo.


clalancette (approved):
This looks fine.  Note, however, that we do *not* bloom-release this one, so once you merge and tag this in the source repository, there is no further work to do here.

marcoag:
We should be good with this then: https://github.com/ros2/system_tests/tree/0.20.1

PR #544, opened 2024-04-10
Actually skip the gtest_subscription test on Connext.

clalancette:
After a relatively recent refactoring, the old way we were marking to skip the test no longer worked (because in the new system, there are no targets that are actually named gtest_subscription__rmw_connextdds).  Use another method to skip the test.

### DIFF
--- a/test_rclcpp/CMakeLists.txt
+++ b/test_rclcpp/CMakeLists.txt
@@ -366,6 +366,9 @@ if(BUILD_TESTING)
         ${rmw_implementation_env_var}
     )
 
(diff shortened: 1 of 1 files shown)


clalancette:
Related to https://github.com/ros2/rmw_connextdds/issues/136

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20815/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15119/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21500/)

ahcorde (approved):

PR #550, opened 2024-10-04
Fix zenoh shutdown

ahcorde:
Fix zenoh shutdown

Related with https://github.com/ros2/rmw_zenoh/pull/276

### DIFF
--- a/test_communication/test/action_server_py.py
+++ b/test_communication/test/action_server_py.py
@@ -153,21 +153,22 @@ if __name__ == '__main__':
         parser.add_argument('namespace', help='namespace of the ROS node')
         args = parser.parse_args()
 
-        with rclpy.init(args=[]):
-            node = rclpy.create_node('action_server', namespace=args.namespace)
+        rclpy.init(args=[])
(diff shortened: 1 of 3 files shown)


test_communication/test/action_server_py.py
```diff
@@ -153,21 +153,22 @@ def execute_goal(goal_handle):
...
-                    node,
-                    NestedMessage,
-                    generate_expected_nested_message_goals(),
-                )
-            else:
-                print('Unknown action type {!r}'.format(args.action_type), file=sys.stderr)
-                sys.exit(1)
-
-            rclpy.spin(node)
+        rclpy.init(args=[])
```
fujitatomoya, line 156: what happens with zenoh if we use contextmanager manner with `InitContextManager`? i am not sure if this change is needed here because both call cleanup process through the context.

clalancette: Yes, exactly, this should not be needed.  Going forward, our recommended pattern in rclpy is to use the context manager with `rclpy.init`, so if that is not working, there is a bug somewhere that needs to be fixed.

PR #552, opened 2024-10-04
Fixed zenoh cli mapping error

ahcorde:
Fixed zenoh cli mapping error

Related to https://github.com/ros2/rmw_zenoh/issues/286

### DIFF
--- a/test_cli_remapping/test/name_maker.py
+++ b/test_cli_remapping/test/name_maker.py
@@ -37,8 +37,9 @@ class NameMaker(rclpy.node.Node):
 
 if __name__ == '__main__':
     try:
-        with rclpy.init():
-            node = NameMaker()
-            rclpy.spin(node)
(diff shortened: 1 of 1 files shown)


test_cli_remapping/test/name_maker.py
```diff
@@ -37,8 +37,9 @@ def __init__(self):
 
 if __name__ == '__main__':
     try:
-        with rclpy.init():
-            node = NameMaker()
-            rclpy.spin(node)
+        rclpy.init()
+        node = NameMaker()
+        rclpy.spin(node)
+        rclpy.shutdown()
```
clalancette, line 43: This shouldn't be necessary; leaving the `rclpy.init` context should automatically call the shutdown.  If that isn't working, then there is a bug somewhere else.

clalancette (changes requested):

Issue #559, opened 2025-02-24
Skipping multi-RMW tests defined in test_communication for rmw_zenoh 


Yadunund:
As we look to run CI with `rmw_zenoh`, a non-DDS based middleware which would not support any cross-vendor tests, we'll need a way to skip running multi-RMW tests part of `test_communication`. At present `system_tests/test_communication` sets [SKIP_MULTI_RMW_TESTS to OFF](https://github.com/ros2/system_tests/blob/be3e512dd0edc033ce877898ea8b22220dcebc50/test_communication/CMakeLists.txt#L21-L22) by default while [explictly excluding certain cross-vendor tests](https://github.com/ros2/system_tests/blob/be3e512dd0edc033ce877898ea8b22220dcebc50/test_communication/CMakeLists.txt#L164-L220) for certain DDS middleware combinations.

If we add `rmw_zenoh` to the `ros2.repos` file and run CI, we can expect failures across all cros-vendor tests. 

One "easy" way to resolve this is to add further exceptions to the `CMakeLists.txt` linked above for `rmw_zenoh`.

Another suggestion from @cottsay is to be able to infer "groups" of RMWs that are expected to be compatible on a per-test basis.

Related to https://github.com/ros2/system_tests/issues/429.

PR #560, opened 2025-02-25
Skip all multi-vendor pub/sub tests with zenoh

Yadunund:
Address #559.

> Note: we already skip all multi-vendor tests between other RMWs for services & actions.

### DIFF
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -195,6 +195,15 @@ if(BUILD_TESTING)
       set(rmw_implementation2_is_cyclonedds TRUE)
     endif()
 
+    set(rmw_implementation1_is_zenoh FALSE)
+    set(rmw_implementation2_is_zenoh FALSE)
+    if(rmw_implementation1 MATCHES "(.*)zenoh(.*)")
+      set(rmw_implementation1_is_zenoh TRUE)
+    endif()
+    if(rmw_implementation2 MATCHES "(.*)zenoh(.*)")
+      set(rmw_implementation2_is_zenoh TRUE)
+    endif()
+
     set(PUBLISHER_RMW ${rmw_implementation1})
     set(SUBSCRIBER_RMW ${rmw_implementation2})
     set(TEST_MESSAGE_TYPES "")
@@ -217,6 +226,14 @@ if(BUILD_TESTING)
           (rmw_implementation1_is_connext AND rmw_implementation2_is_cyclonedds) OR
           (rmw_implementation1_is_cyclonedds AND rmw_implementation2_is_connext)
         )
+      )
+        continue()
+      endif()
+      # Zenoh will not interoperate with any DDS based middlewares so we skip
+      # all multi-vendor pub/sub tests.
+      if(
+        (rmw_implementation1_is_zenoh AND NOT rmw_implementation2_is_zenoh) OR
+        (rmw_implementation2_is_zenoh AND NOT rmw_implementation1_is_zenoh)
       )
         continue()
       endif()
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -235,7 +235,7 @@ if(BUILD_TESTING)
         (rmw_implementation1_is_zenoh AND NOT rmw_implementation2_is_zenoh) OR
         (rmw_implementation2_is_zenoh AND NOT rmw_implementation1_is_zenoh)
       )
-        continue()
+        set(SKIP_TEST "SKIP_TEST")
       endif()
       list(APPEND TEST_MESSAGE_TYPES "${message_type}")
     endforeach()
--- a/test_communication/CMakeLists.txt
+++ b/test_communication/CMakeLists.txt
@@ -203,6 +203,14 @@ if(BUILD_TESTING)
     if(rmw_implementation2 MATCHES "(.*)zenoh(.*)")
       set(rmw_implementation2_is_zenoh TRUE)
     endif()
+      # Zenoh will not interoperate with any DDS based middlewares so we skip
+      # all multi-vendor tests.
+    if(
+      (rmw_implementation1_is_zenoh AND NOT rmw_implementation2_is_zenoh) OR
+      (rmw_implementation2_is_zenoh AND NOT rmw_implementation1_is_zenoh)
+    )
+      set(SKIP_TEST "SKIP_TEST")
+    endif()
 
     set(PUBLISHER_RMW ${rmw_implementation1})
     set(SUBSCRIBER_RMW ${rmw_implementation2})
@@ -228,14 +236,6 @@ if(BUILD_TESTING)
         )
       )
         continue()
-      endif()
-      # Zenoh will not interoperate with any DDS based middlewares so we skip
-      # all multi-vendor pub/sub tests.
-      if(
-        (rmw_implementation1_is_zenoh AND NOT rmw_implementation2_is_zenoh) OR
-        (rmw_implementation2_is_zenoh AND NOT rmw_implementation1_is_zenoh)
-      )
-        set(SKIP_TEST "SKIP_TEST")
       endif()
       list(APPEND TEST_MESSAGE_TYPES "${message_type}")
     endforeach()


cottsay:
I'm not 100% certain about the mechanics at play here, but I think it would be more correct to restore the previous behavior using the call to `continue()`. Skipping a test carries some implication that the test is valid but we're choosing not to run it, where I think in this case the test isn't valid - we'd never want to attempt this sort of communication. If I'm reading the CMake correctly, the call to `continue()` results in no test being created at all, which I think is what we want.

Yadunund:
> I'm not 100% certain about the mechanics at play here, but I think it would be more correct to restore the previous behavior using the call to `continue()`. Skipping a test carries some implication that the test is valid but we're choosing not to run it, where I think in this case the test isn't valid - we'd never want to attempt this sort of communication. If I'm reading the CMake correctly, the call to `continue()` results in no test being created at all, which I think is what we want.

I had assumed the same but with `continue()`, I still saw cross-vendor tests being generated with `rmw_zenoh` in my local testing. Will look closer at what's going on.

test_communication/CMakeLists.txt
```diff
@@ -220,6 +229,14 @@ if(BUILD_TESTING)
       )
         continue()
       endif()
+      # Zenoh will not interoperate with any DDS based middlewares so we skip
+      # all multi-vendor pub/sub tests.
+      if(
+        (rmw_implementation1_is_zenoh AND NOT rmw_implementation2_is_zenoh) OR
+        (rmw_implementation2_is_zenoh AND NOT rmw_implementation1_is_zenoh)
+      )
+        set(SKIP_TEST "SKIP_TEST")
```
fujitatomoya: i think if we call `continue()` here, it will start the test with `add_launch_test` with timeout 0. it seems strange that there is no message type `WStrings.msg` either in "rosidl_interfaces" or "test_msgs"... maybe in the past, there were cases to skip those cases?

i would suggest that we can call `set(SKIP_TEST "SKIP_TEST")` line 206 right after validating which rmws are going to be tested? we do not have to check that in the foreach loop of message types, which are irrelevant from the criteria?

Yadunund: > i think if we call continue() here, it will start the test with add_launch_test with timeout 0.

Thanks for providing that insight. Instead of timing out, wouldn't it be better to skip this test entirely with set(SKIP_TEST "SKIP_TEST") as I have it now?

> it seems strange that there is no message type WStrings.msg either in "rosidl_interfaces" or "test_msgs"... maybe in the past, there were cases to skip those cases?

My understanding is that all rosidls generate the WString message by default and without a WString.msg file defined in test_msgs. See https://github.com/ros2/rosidl/issues/352. This message is then included [here in test_communication](https://github.com/ros2/system_tests/blob/be3e512dd0edc033ce877898ea8b22220dcebc50/test_communication/test/subscribe_string_types.hpp#L23). I've checked my local workspace and indeed there is such a header file present.  So I don't think we want to skip these tests entirely...

fujitatomoya: i see, that makes more sense. thanks for sharing the information.

fujitatomoya (approved):
lgtm with green CI

test_communication/CMakeLists.txt
Yadunund: Actually with `continue()`, I'm seeing the test timeout being generated for 60s 🤔 

```bash
yadunund@ubuntu-24-04:~/ros2_rolling_ci$ colcon test --event-handlers console_cohesion+ console_direct+ --packages-select test_communication --ctest-args -R test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp
Starting >>> test_communication
UpdateCTestConfiguration  from :/home/yadunund/ros2_rolling_ci/build/test_communication/CTestConfiguration.ini
Parse Config file:/home/yadunund/ros2_rolling_ci/build/test_communication/CTestConfiguration.ini
   Build name: (empty)
 Add coverage exclude regular expressions.
Create new tag: 20250303-2050 - Experimental
UpdateCTestConfiguration  from :/home/yadunund/ros2_rolling_ci/build/test_communication/CTestConfiguration.ini
Parse Config file:/home/yadunund/ros2_rolling_ci/build/test_communication/CTestConfiguration.ini
Test project /home/yadunund/ros2_rolling_ci/build/test_communication
Constructing a list of tests
Done constructing a list of tests                
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 242
    Start 242: test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp

242: Test command: /usr/bin/python3 "-u" "/home/yadunund/ros2_rolling_ci/install/ament_cmake_test/share/ament_cmake_test/cmake/run_test.py" "/home/yadunund/ros2_rolling_ci/build/test_communication/test_results/test_communication/test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp.xunit.xml" "--package-name" "test_communication" "--output-file" "/home/yadunund/ros2_rolling_ci/build/test_communication/launch_test/test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp.txt" "--env" "PYTHONDONTWRITEBYTECODE=1" "--append-env" "LD_LIBRARY_PATH=/home/yadunund/ros2_rolling_ci/build/test_communication" "--command" "/usr/bin/python3" "-m" "launch_testing.launch_test" "/home/yadunund/ros2_rolling_ci/build/test_communication/test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp_Release.py" "--junit-xml=/home/yadunund/ros2_rolling_ci/build/test_communication/test_results/test_communication/test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp.xunit.xml" "--package-name=test_communication"
242: Working Directory: /home/yadunund/ros2_rolling_ci/build/test_communication
242: Test timeout computed to be: 60
242: -- run_test.py: extra environment variables:
242:  - PYTHONDONTWRITEBYTECODE=1
242: -- run_test.py: extra environment variables to append:
242:  - LD_LIBRARY_PATH+=/home/yadunund/ros2_rolling_ci/build/test_communication
242: -- run_test.py: invoking following command in '/home/yadunund/ros2_rolling_ci/build/test_communication':
242:  - /usr/bin/python3 -m launch_testing.launch_test /home/yadunund/ros2_rolling_ci/build/test_communication/test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp_Release.py --junit-xml=/home/yadunund/ros2_rolling_ci/build/test_communication/test_results/test_communication/test_publisher_subscriber__rclcpp__rmw_zenoh_cpp__rmw_fastrtps_dynamic_cpp.xunit.xml --package-name=test_communication
242: [INFO] [launch]: Default logging verbosity is set to INFO
242: [INFO] [daemon-stop-1]: process started with pid [102325]
242: [INFO] [daemon-stop-1]: process has finished cleanly [pid 102325]
242: test_subscriber_terminates_in_a_finite_amount_of_time[] (test_communication.TestPublisherSubscriber.test_subscriber_terminates_in_a_finite_amount_of_time[])
242: Test that the subscriber terminates after a finite amount of time. ... [INFO] [test_publisher-2]: process started with pid [102362]
242: [INFO] [test_subscriber-3]: process started with pid [102363]
242: [test_publisher-2] Unknown message argument ''
242: [ERROR] [test_publisher-2]: process has died [pid 102362, exit code 1, cmd '/home/yadunund/ros2_rolling_ci/build/test_communication/test_publisher_cpp /test_time_20_50_15'].
242: [test_subscriber-3] Unknown message argument ''
242: [ERROR] [test_subscriber-3]: process has died [pid 102363, exit code 1, cmd '/home/yadunund/ros2_rolling_ci/build/test_communication/test_subscriber_cpp /test_time_20_50_15'].
242: ok
```

mjcarroll: Just gaining context here, but I guess this is unexpected behavior?  If the test is skipped, why are we still getting a 60 second timeout?

wjwwood: Just looking at how we skip pub/sub tests above this, maybe we should just add a block like on lines 173-178?

`continue()` here is the only thing that makes sense, as this loop is responsible for building up the messages being tested and then use that to calculate a timeout. SKIP_TEST here doesn't really make sense, as it's not taking affect on a per message basis anyway.

> So I don't think we want to skip these tests entirely...

I think that's exactly what we want to do with the logic you implemented, just higher that this, out of the loop, like @fujitatomoya suggested. 

I'll make the change and see that it runs the zenoh-zenoh tests, but not other cross vendor tests with zenoh.

Yadunund: @wjwwood nice find. I've pushed https://github.com/ros2/system_tests/pull/560/commits/60d45d5f625109ec1a212ec5bec36e2184e584f9 as per your recommendations. Tested locally and all multi-vendor tests are skipped while zenoh-zenoh tests run.

Yadunund:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22718/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16955/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23582/)

test_communication/CMakeLists.txt
wjwwood: You beat me to it, I was about to test it locally.

wjwwood (approved):

ahcorde:
https://github.com/Mergifyio backport jazzy

